PR #39, opened 2023-07-27
[#11] Add lib to create background job using BullMQ

### LINKED ISSUE #11: [Backend] Set a background job to run the scraping process for Google search
## Why

As the requirement of the project is to scrape the google as soon as the file is uploaded. So, to run the process automatically we are setting up a job to trigger the scraping of result automatically. 

## Acceptance Criteria

- Setup a BullMQ job once the file is uploaded

## Implementation Details

- Setup a BullMQ and required packages for it
- Make a separate class that can add job to BullMQ

Prabeshpd:
- Close #11 

## What happened 👀

- [x] Created a lib to create a background job 
- [x] Added BullMQ package
- [x] Add tests

## Insight 📝

Scraping is a time-consuming task, so it must be segregated from regular requests to the backend. Separate jobs are created and executed with the help of BullMQ. 

I created an interface for queue and worker implementation. The reason behind this was to make the code as much as independent of the package. So, in the future, if we have a separate package and implementation, I want to bind them on a certain interface so the code that is using the service won't get changed. 

## Proof Of Work 📹

Screen Shot 2566-07-31 at 09 39 42


tednguyendev (commented):
LGTM!

You can also add a script into the file `package.json` so that we can do something like `yarn worker` to run the worker, too :D

And we should add that script ^ into the README also 🙈

src/utils/worker.test.ts
```diff
@@ -0,0 +1,36 @@
...
+  });
+
+  it('adds new job on calling the add implementation for BullQueue class', async () => {
+    const bullMq = new BullQueue(mockQueue);
+    await bullMq.addJob({ name: 'asfd', options: {}, data: {} });
+
+    mockQueue.add.calledOnceWith('asdf', {}, {});
+  });
+
+  it('add multiple jobs on calling the bulk add for BullQueue', async () => {
```
tednguyendev, line 23: Mini typo: should be "adds" 🙈

src/utils/worker.test.ts
```diff
@@ -0,0 +1,36 @@
+/**
+ * @jest-environment node
+ */
+
+import { Queue } from 'bullmq';
+import sinon from 'sinon';
+
+import { BullQueue } from '../utils/worker';
+
+describe('BullQueue():', () => {
```
tednguyendev, line 10: I haven't worked with Javascript much yet, so it's just me being curious here 👀 : Is this the convention that we usually use when we need to test a class in Javascript? 🤔 

It's just that, in Ruby, we will only do something like 
```ruby
describe('BullQueue')
```

Instead of 
```javascript
describe('BullQueue():')
```
(Not have the `():`)

Thank you in advance :D

src/utils/worker.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+interface IQueue<T> {
+  addJob: (job: IJob) => void;
+  queue: T;
+  addBulkJobs: (job: IJob[]) => void;
+  clearJobs: () => void;
+}
+
+export class BullQueue implements IQueue<Queue> {
```
tednguyendev, line 23: I like the way you have separate interfaces for queues and workers too :D

tednguyendev:
Oh, and I think adding the "Milestone" into a PR is enough,

CleanShot 2023-07-31 at 10 27 16@2x

We don't need to also add a lable for that milestone, although having it won't hurt 😄 

CleanShot 2023-07-31 at 10 27 45@2x

src/utils/worker.test.ts
Prabeshpd: It was just a convention in my old office. Normally we do not add the parenthesis. Old habit 😅

tednguyendev: Got it. Thank you!

src/services/scraper.ts
olivierobert, line 1: Why adding this empty file?

src/utils/worker.test.ts
```diff
@@ -0,0 +1,36 @@
...
+  let mockQueue: sinon.SinonStubbedInstance<Queue>;
```
olivierobert, line 11: Why using `sinon` over standard [Jest Mocks](https://jestjs.io/docs/mock-functions). The fewer dependencies for testing, the better.

olivierobert (commented):
`utils` and `helpers` can become the place where we put code when we do not know where to put it, hence becoming messy. Prefer using the directory `src/lib` as it refers to local libraries (in opposition to third-party libraries living in `node_module`).

Prabeshpd:
> `utils` and `helpers` can become the place where we put code when we do not know where to put it, hence becoming messy. Prefer using the directory `src.lib` as it refers to local libraries (in opposition to rhid-party libraries living in `node_module`).

Ok. 😄

src/services/scraper.ts
Prabeshpd: It is for the scraping logic. Will be in next PR.Will delete this for now.

Prabeshpd:
> LGTM!
> 
> You can also add a script into the file `package.json` so that we can do something like `yarn worker` to run the worker, too :D
> 
> And we should add that script ^ into the README also 🙈

Since this is only the implementation as a library to be implemented. Will add the script in next PR for running the worker

tednguyendev (approved):

olivierobert (approved):

olivierobert:
I have updated the PR title and description to reflect the current implementation (moving from `util` to `lib`). Moving forward ensure you update these details.

Issue #4, opened 2023-07-17
[Chore] Set up DB for the project


Prabeshpd:
## Why

When running the application user may want to view the activity they have performed in the past for it we would need the persistent data. For this we want to leverage PostgreSQL for the storing the data

## Acceptance Criteria

- Setup connection mechanism to connect to a database instance based on the given connection parameter (From external application such as Vault or from the env file)

## Implementation Details

- Use Vercel Postgres for development
- Use Prisma for facilitating migration
- Use env file to store the credentials

## Resource
[Guide](
https://vercel.com/guides/nextjs-prisma-postgres)

Issue #3, opened 2023-07-17
[Chore] Setup CI environment to run the test


Prabeshpd:
## Why

Whenever a new PR is merged or created we need to check whether all the codes integrates properly or not. To test it one must create an environment and run the test written. For this we want to leverage the GitHub Actions for running the test and building the application

## Acceptance Criteria

- Create a CI pipeline that runs on PR running all the test

## Implementation Details
- Create CI so that the end to end test can be run
- Create CI so that the unit test can be run

olivierobert:
@Prabeshpd I am unsure what the difference is between this chore and #2 . Both could be merged into one.

Issue #2, opened 2023-07-17
[Chore] Set up CI pipeline for the project


Prabeshpd:
## Why
To build the application we would need require few dependency such as nextJs, Typescript and few testing frameworks to facilitate the development process

## Acceptance Criteria

- Add the required dependencies with nextJS and testing frameworks

## Implementation Details
- Add CI pipeline for validating the linting with eslint and prettier
- Add CI pipeline to validate the build of the project
- Add CI pipeline to validate docker image build
- Add CI pipeline for test

Issue #1, opened 2023-07-17
[Backend] As a user, I can sign up with email and password


Prabeshpd:
## Why

User must be authenticated to use certain features of application. To be authenticated one must be able to register themselves to application.

## Acceptance Criteria

- User must be able to register themselves with username and password
- Password must be encrypted before saving
- Email must be unique
- Proper message must be send to user that registration is completed
- Should receive a message on dashboard that they have successfully registered.
- Should have test configured

PR #43, opened 2023-07-31
[#14] [Backend] As a user, my uploaded keywords are scraped and saved once uploaded

### LINKED ISSUE #14: Scrape the keywords and store it in database
## Why

Once the job us setup to scrape a separate process must get the result from google search, scrape it and save it in the database for the client to view the results.

## Acceptance Criteria
Search the google with keyword
Scrape the required results
Store them in database
Place a fallback mechanism to retry if the process fails and after n number of retries log it and move to other keywords

Prabeshpd:
- Close #14 

## What happened 👀

- [x] Add a bullmq job to scrape google search and store it in database
- [x] Add a functionality to search the google search
- [x] Add a functionality to parse the google search
- [x] Add a delayer when the job is added for the scraper job
- [x] Add github actions for the database migrations when PR is merged to staging and production
- [x] Add tests

## Insight 📝

Once the key words are stored we need to add a job to scrape the result and store them in database along with the update on statuses whether the scrape was successful or not. For this bullmq job is used and bullmq delayed job is leveraged to delay the processing of each jobs.

> [!NOTE] 
 > For this no integration test is added on this PR as there will be a separate PR for it along with CI pipelines to run it.

## Proof Of Work 📹

Screen Shot 2566-08-07 at 14 37 49


.github/workflows/migrate_production.yml
```diff
@@ -0,0 +1,33 @@
+name: Migrate Database Schema To Production
```
olivierobert, line 1: This should be an extra step in the exsting workflows for `deploy_staging` and `deploy_production`. Just like building the JS files, migrating the DB is part of the deployment workflow.

src/constants/enums.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export enum ERROR_TYPES {
+  NOT_FOUND = 'NOT_FOUND_ERROR',
+  BAD_REQUEST = 'BAD_REQUEST_ERROR',
+  UNAUTHORIZED = 'UNAUTHORIZED',
+  INTERNAL_SERVER_ERROR = 'INTERNAL_SERVER_ERROR',
+  INVALID_PAYLOAD = 'INVALID_PAYLOAD',
+}
+
+export const ERROR_CODES = new Map<ERROR_TYPES, number>([
```
olivierobert, line 11: Technically this is not an enum so the file might better be named `error.ts`. Also avoid plural whenever possible: https://nimblehq.co/compass/development/code-conventions/#convention-15

src/lib/worker.test.ts
```diff
@@ -25,16 +25,16 @@ describe('BullQueue():', () => {
 
   it('adds new job on calling the add implementation for BullQueue class', async () => {
     const bullMq = new BullQueue(mockQueue);
-    await bullMq.addJob({ name: 'asfd', options: {}, data: {} });
+    await bullMq.addJob({ name: 'asfd', opts: {}, data: {} });
```
olivierobert, line 28: `options` was better because abbreviations should be avoided: https://nimblehq.co/compass/development/code-conventions/#convention-12

src/workers/buildScraper.ts
```diff
@@ -0,0 +1,29 @@
...
+  userId: number;
+  keyword: string;
+}
+
+const buildScraper = async (job: Job) => {
+  const { userId } = job.data;
+
+  const searchResults = await listSearchResult(userId);
+
+  const delayer = new Delayer<IScraperData>(
```
olivierobert, line 18: The whole delying mechanism is too complex 😵 A simpler solution would be to use the `delay` option (https://docs.bullmq.io/guide/jobs/delayed) and calculate an exponentail delay value, e.g.  job #1 -> delay is 0, job #2 -> delay is 2s, job #3 -> delay is 4s, job #4 -> delay is 12s, etc. 

> [!Note]
> This is called exponental backoff: https://en.wikipedia.org/wiki/Exponential_backoff 🧑‍🏫

olivierobert (changes requested):
All files in `utils` should be moved to `lib`. It is best to have `lib` and `helpers` directories.

This PR is doin too much, thus lenghtneing the implementation time and making the code review harder 🤯 It is best to have separate chores to add the migration workflows. improved the Logger,  and even for the feature for exponential backoff (it is not required if we scrape only one keyword). 

> [!Warning]
> Prefer to adopt an iterative process with small pieces of work merged regularly.

src/lib/worker.test.ts
Prabeshpd: BullMq jobs interface has native opts keyword. So, went with the opts keyword

src/workers/buildScraper.ts
Prabeshpd: 😅 I was doing the same thing as the delay option in bullmq but in more complicated way. Will use bullmq delayer.

Prabeshpd:
> All files in `utils` should be moved to `lib`. It is best to have `lib` and `helpers` directories.
> 
> This PR is doin too much, thus lenghtneing the implementation time and making the code review harder 🤯 It is best to have separate chores to add the migration workflows. improved the Logger, and even for the feature for exponential backoff (it is not required if we scrape only one keyword).
> 
> Warning Prefer to adopt an iterative process with small pieces of work merged regularly.

Ok. Noted. Was planning to break the PR once the feature was completed and then make separate issue for each PR. But makes sense to separate them beforehand and plan them more properly. Will improve this on the upcoming tasks 🙂

src/workers/buildScraper.ts
Prabeshpd: Fixed in [038e5ab](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/43/commits/038e5abc75968d27c81099ae677e152a17b09cfd)

src/services/parser.test.ts
```diff
@@ -0,0 +1,76 @@
...
+    adTopParent: '#tvcap',
+    adBottomParent: '#bottomads',
+    adWords: '.uEierd',
+    nonAdWords: '.MjjYud',
+    linkElement: 'a',
+    linkElementAttribute: 'href',
+    nonAdParent: '#search',
+  };
+  it('returns top AdWord list count if the matching attribute is formed', () => {
+    const rawHtmlString = `
```
tednguyendev, line 21: This can be extracted to a separate HTML file.

Doing that will make the code here looks more neat, plus, as we put them in an HTML file, the text editors will apply syntax highlighting to the text in HTML files, making it looks even better! 👐

src/workers/scraper.ts
```diff
@@ -1,6 +1,45 @@
...
-export const scrapeQueue = new BullQueue(scrapeBullQueue);
+const { selectorIdsSearch } = config;
+
+export const scrape = async (job: Job) => {
+  try {
+    const { keyword } = job.data;
+    const htmlData = await searchResult(keyword);
+    const parserInstance = new Parser(selectorIdsSearch, htmlData);
+    const htmlDocumentAPI = parserInstance.loadHtmlDocumentApi();
+    const topAdParsed = parserInstance.parseTopAdWordList(htmlDocumentAPI);
```
tednguyendev, line 16: Do you thing the action of passing `htmlDocumentAPI` into the parse functions a little bit repetitive? 🤔 

I don't know if this is the Javascript way, but can we just save the `htmlDocumentAPI` into the Parser instance when we calls `loadHtmlDocumentApi` like this?

```js
class Parse implements IParse {
  htmlDocumentAPI: any;

  loadHtmlDocumentApi() {
    this.htmlDocumentAPI = cheerio.load(this.htmlRawString);
  }
}
```

Then we can just:
```js
const topAdParsed = parserInstance.parseTopAdWordList();
const bottomAdParsed = parserInstance.parseBodyAdWordList();
const nonAdParsed = parserInstance.parseNonAdWordList();
```

src/workers/scraper.ts
```diff
@@ -1,6 +1,45 @@
...
+    const bottomAdParsed = parserInstance.parseBodyAdWordList(htmlDocumentAPI);
+    const nonAdParsed = parserInstance.parseNonAdWordList(htmlDocumentAPI);
+    const adWordsCount = topAdParsed.count + bottomAdParsed.count;
+    const nodAdUrls = nonAdParsed.urls;
+    const args = {
+      where: {
+        userId: 1,
```
tednguyendev, line 23: Why are we hardcoding the `userId`? 👀

tednguyendev: You can always add some comments or use the PR description to explain things that look strange but has their hidden meanings before opening a PR ✋

src/services/parser.ts
```diff
@@ -0,0 +1,85 @@
...
+    this.selectorIds = selectorIds;
+    this.htmlRawString = htmlRawString;
+  }
+
+  loadHtmlDocumentApi() {
+    return cheerio.load(this.htmlRawString);
+  }
+
+  parseTopAdWordList(htmlDocumentAPI: cheerio.CheerioAPI) {
+    const urls: string[] = [];
```
tednguyendev, line 32: A mini tip I use to make code clean is to only declare variables only when I need them.

Yes, this function needs the `urls`, but if we declare it on line 40(above the `adUrlElements.each`, which uses `urls`), that will reduce the mild confusion when the developers move their eyes from line 32 to 43, wondering why the code still not use this `urls` 😆

src/workers/scraper.ts
Prabeshpd: Yes, the initial thought was to do this. Since, typescript doesn't track this mutation in that case and it makes us compulsory to initialize the property in constructor. To mitigate that I have to either use IIFE function or make them any type or add null type to initialize it to value. Again coming to parse function I have to check in every instance whether the property has value or not which adds more repetitive code. So, have gone for this approach. Since here, we only have to pass variable to function. If the code would be repetitive would have gone with the above approach.

src/services/parser.ts
Prabeshpd: 😄 . It is a javaScript convention to declare variables at the top. As per my knowledge before the use of let and const the variables are not blocked scoped so defining variable down would hoist them and sometime may leave the variables undeclared so, was made convention to declare variables in the top. Since, let and const as they are block scoped we can declare them later.

src/services/parser.test.ts
Prabeshpd: If a separate file is declared then again I have to add other helper function to read and parse the file while importing it. Since, the raw string are small and maintained proper spacing I had opted for directly defining them here. If the file are big and needed much more frequently would have gone for that approach.

src/workers/scraper.ts
Prabeshpd: OOPS. 🙈  Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/43/commits/ee0896e5344d5362ad2487d35aa77e50395d7210.

src/services/parser.ts
tednguyendev: I see, so that's another reason why people use `var` these days, right? 🤔 Thanks for your explanation.

Prabeshpd: I think you meant that people `don't` use `var`😅  right?

src/workers/scraper.ts
tednguyendev: > it makes us compulsory to initialize the property in constructor.

So you means that, by doing this, it will work? 👀
```js
class Parse implements IParse {
  selectorIds: ISelectorId;
  htmlRawString: string;
  htmlDocumentAPI: any;

  constructor(selectorIds: ISelectorId, htmlRawString: string) {
    this.selectorIds = selectorIds;
    this.htmlRawString = htmlRawString;
    this.htmlDocumentAPI = cheerio.load(this.htmlRawString);
  }
}
```

Can we do that? 🤔

Prabeshpd: yes, we can do that. 😄 . I was just trying to make the class independent of library. But thinking out loud it just makes the interface dependent on the library 😅. Will go for this approach.

Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/43/commits/2b704b3a699327a3742c07232e567214524b4ca1

src/services/parser.ts
tednguyendev: Ah yes 🤣

tednguyendev (approved):

src/workers/scraper.test.ts
```diff
@@ -0,0 +1,55 @@
...
+
+describe('Scrape:', () => {
+  let mockJob: Partial<Job>;
+  beforeEach(() => {
+    mockJob = {};
+    mockClear(Parse.prototype.loadHtmlDocumentApi);
+    mockClear(Parse.prototype.parseBodyAdWordList);
+    mockClear(Parse.prototype.parseNonAdWordList);
+    mockClear(Parse.prototype.parseTopAdWordList);
+    mockClear(search.searchResult);
```
olivierobert, line 27: Doesn't `jest.resetAllMocks` or `jest.restoreAllMocks` do the same?

src/workers/scraper.test.ts
```diff
@@ -0,0 +1,55 @@
...
+  });
+
+  it('calls all the functions one time for scraping process', async () => {
+    mockJob.data = { keyword: 'node' };
+    mockedSearchResult.searchResult.mockResolvedValue('asdfdf');
+
+    await scrape(mockJob as Job);
+
+    expect(search.searchResult).toHaveBeenCalledTimes(1);
+    expect(Parse.prototype.parseBodyAdWordList).toHaveBeenCalledTimes(1);
```
olivierobert, line 37: Why do we need to spy on the prototype method? Wouldn't https://jestjs.io/docs/jest-object#jestspyonobject-methodname work? The mway the mock object was set `search as jest.Mocked<typeof search>;` is rather odd/unexpected.

olivierobert (changes requested):
This pull request is very large and there is too many implementation details. So it will take time to have it merged 😓

PR #44, opened 2023-08-02
[#45] [Backend] As a user I can perform database operation

### LINKED ISSUE #45: [Backend] Implement database operation
## Why

For application to run application may need to save the persistent data to database. So, it is requirement for the application to have basic database functionalities connecting to a database.

## Acceptance Criteria

- Make a  connection to database using Prisma
- Create migration files
- Add functions to connect to specific tables

Prabeshpd:
- Close #45 

## What happened 👀

- [x] Added the functionality to create user
- [x] Added the functionality to create search result with keywords and update it
- [x] Added the functionality to retrieve search result based on provided userId
- [x] Add the tests

## Insight 📝

For scraping the results one must save the keywords with the user_id for which user the keywords are stored persistently. To solve this a new migration for search result is created along with the functionalities to save, update and retrieve them.

## Proof Of Work 📹

Screen Shot 2566-08-03 at 16 32 46


olivierobert:
@Prabeshpd  The PR title must match the Issue: "[Backend] As a user, my uploaded keywords are scraped and saved once uploaded". The process is to have one branch + PR for each story.

Prabeshpd:
> @Prabeshpd The PR title must match the Issue: "[Backend] As a user, my uploaded keywords are scraped and saved once uploaded". The process is to have one branch + PR for each story.
> 
> Note I fixed the merged PRs to follow this convention as us - code reviewers - failed to notice this issue previously 🙈

Ok. 😄 Will create a separate issue for this one. Since this came while I was working is it ok to create a new issue and add it to this sprint? If I put in the same PR the PR will have lot of changes 🙈 . It seems I underestimated this issue to further break it down

prisma/schema.prisma
```diff
@@ -4,17 +4,38 @@ generator client {
...
+  url      = env("POSTGRES_DB_CONNECTION_URL")
 }
 
 model User {
-  id          String       @id @default(uuid())
-  name        String       @db.VarChar(100)
-  email       String       @db.VarChar(50)
-  createdAt   DateTime     @default(now()) @map("created_at")
-  updatedAt   DateTime     @default(now()) @map("updated_at") @updatedAt
+  id           Int            @id @default(autoincrement())
```
tednguyendev, line 11: Why should we change it from `String` to `Int`? 👀

prisma/migrations/20230803044907_camel_case/migration.sql
```diff
@@ -0,0 +1,34 @@
...
+  - You are about to drop the column `user_id` on the `search_results` table. All the data in the column will be lost.
+  - Added the required column `userId` to the `search_results` table without a default value. This is not possible if the table is not empty.
+
+*/
+-- DropForeignKey
+ALTER TABLE "search_results" DROP CONSTRAINT "search_results_user_id_fkey";
+
+-- AlterTable
+ALTER TABLE "search_results" DROP COLUMN "ad_word_top_urls",
+DROP COLUMN "ad_words_count",
```
tednguyendev, line 19: I think we should use only 1 migration file, which will create both `search_results` table and use the camel case fields instead of using 2 migration files(this file and [the previous file](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/44/files#diff-bdb9ea9dbfc956baeb941e46d8cd9a6b0601822efbd5a9b7c3e6c0befac944d7R1-R21)).

src/lib/crypt.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export async function hash(value: string): Promise<string> {
+  const saltRounds = parseInt(config.auth.saltRounds, 10);
+  const hashedValue = await bcrypt.hash(value, saltRounds);
+
+  return hashedValue;
+}
+
+export async function compare(value: string, hashedValue: string): Promise<boolean> {
+  const isSame = await bcrypt.compare(value, hashedValue);
```
tednguyendev, line 15: How about just return the `await bcrypt.compare(value, hashedValue)` without using any variables like `isSame` 🙈

tednguyendev: And the `compare` function will be used for the next PR right? 👀

tednguyendev:
Ah, and I just realized that maybe the PR does not have enough tests yet, although the status of the PR is in review, and the the box `Add the tests` is ticked as done in the PR. So I will come back later and do more reviews! ✋

Prabeshpd:
> Ah, and I just realized that maybe the PR does not have enough tests yet, although the status of the PR is in review and the box `Add the tests` in the PR description is ticked as done. So I will come back later and do more reviews! ✋

Test are still left. I am adding the test on search Results too. It seems I have mistakenly ticked the add test task

src/lib/crypt.ts
Prabeshpd: Will remove this function. Seems remanent of the rebase done in earlier PR 🙈

prisma/schema.prisma
Prabeshpd: Since, it is a primary field for user. It does not make sense to make it specifically uuid as per now requirement. So, changed it to integer

tednguyendev: I think it's still a normal thing to use UUIDs for primary fields, but yeah, for simple use cases like ours, we can just use id for simplicity. 👌

tednguyendev (approved):
LGTM.

But I think the scope of this PR, as stated on [the issue](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/issues/45), should only be about setting up the database and creating migration files.

The repository files should be added later, when we need to actually use them. Having them early on this PR causes no harm, but without the code that actually uses them, I find it hard to review as a whole flow. For example, [this file](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/44/files#diff-69ac28b9f3b3bdefd2d5040cb90c72b233ff05a82b15545348fcfb0beddaeb7bR6) can be added on the PR to solve [this ticket](https://github.com/orgs/nimblehq/projects/59?pane=issue&itemId=33408557) instead.

Prabeshpd:
> LGTM.
> 
> But I think the scope of this PR, as stated on [the issue](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/issues/45), should only be about setting up the database and creating migration files.
> 
> The repository files should be added later, when we need to actually use them. Having them early on this PR causes no harm, but without the code that actually uses them, I find it hard to review as a whole flow. For example, [this file](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/44/files#diff-69ac28b9f3b3bdefd2d5040cb90c72b233ff05a82b15545348fcfb0beddaeb7bR6) can be added on the PR to solve [this ticket](https://github.com/orgs/nimblehq/projects/59?pane=issue&itemId=33408557) instead.

Yes, It makes sense to place in the PR that we can use. Only thought here was to add the database related file in one PR. I was trying to segregate the scope of PR on the domain that I was working on. But it makes sense to add them on the time of use of these functions. Will keep that in mind on my upcoming PR. Thank you @tednguyendev

prisma/migrations/20230803092725_search_result/migration.sql
olivierobert, line 1: It would be best to split the creation of multiple tables into multiple migrations:
- one migration to create users
- one migration to create search_results

Not a blocker but this is how we would normally do it 😅

src/lib/crypt.test.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * @jest-environment node
+ */
+
+import { hash } from './crypt';
+
+describe('Crypt', () => {
+  describe('hash()', () => {
+    it('should generate hash for the string', async () => {
```
olivierobert, line 9: Avoid the old syntax `should`, prefer the imperative tone:

```suggestion
    it('generates a hash for the string', async () => {
```

src/lib/crypt.test.ts
```diff
@@ -0,0 +1,16 @@
...
+      const str = 'test';
+      const result = await hash(str);
```
olivierobert, line 11: Minor rename suggestion to avoid using too generic variable names such as `data` or `results`:

```suggestion
      const hash = await hash(str);
```

or 

```suggestion
      const generatedHash = await hash(str);
```

src/lib/crypt.test.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * @jest-environment node
+ */
+
+import { hash } from './crypt';
+
+describe('Crypt', () => {
+  describe('hash()', () => {
+    it('should generate hash for the string', async () => {
+      const str = 'test';
```
olivierobert, line 10: Avoid abbreviated variable names (see: https://nimblehq.co/compass/development/code-conventions/#convention-12) and naming after object types (see: https://nimblehq.co/compass/development/code-conventions/#convention-16):

```suggestion
      const inputText = 'test';
```

src/repositories/searchResult.ts
```diff
@@ -0,0 +1,30 @@
...
+}>;
+
+export const createSearchResult = async (
+  searchResultAttributes: Prisma.SearchResultCreateManyInput
+) => dbClient.searchResult.createMany({ data: searchResultAttributes });
+
+export const updateSearchResult = async (args: Prisma.SearchResultUpdateManyArgs) =>
+  dbClient.searchResult.updateMany(args);
+
+export const listSearchResultForScraping = async (userId: number) => {
```
olivierobert, line 21: This method name is a bit inconsistent (the suffix `ForScraping` is added only this method and seems redundant):

```suggestion
export const listSearchResult = async (userId: number) => {
```

olivierobert (changes requested):
Mostly some minor naming issues ✌️

src/repositories/searchResult.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/44/commits/772649044e477fe0425328fbaae9c5acd0eccb73

src/lib/crypt.test.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/44/commits/772649044e477fe0425328fbaae9c5acd0eccb73

src/lib/crypt.test.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/44/commits/772649044e477fe0425328fbaae9c5acd0eccb73

src/lib/crypt.test.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/44/commits/772649044e477fe0425328fbaae9c5acd0eccb73

prisma/migrations/20230803092725_search_result/migration.sql
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/44/commits/08199ffc4324f5b577b4f131651de64d27baac0c

olivierobert (approved):

PR #52, opened 2023-08-07
[#50][Chore] Add integration test for scrape job

### LINKED ISSUE #50: [Chore] Add integration test
## Why

For now there is only mocking of the third party services and dependent functions. To have a complete confidence on the code we would need test integrating all the services or units involved with the code. 

## Acceptance Criteria

- Add integration test for database
- Add integration test for scraping job

Prabeshpd:
- Close #50 

## What happened 👀
Add integration test for the scraping process

## Insight 📝
Unit test mostly mocks the integration of the third party services or database storage. To have complete confidence on our code we would need a test as integration of different services. Here scraping job is tested where 

- the request from axios to get search result
- parsing of cheerion of raw html string
- update of table search_results on completion

are tested.

## Proof Of Work 📹

Screen Shot 2566-08-08 at 10 53 04


olivierobert:
The tests should never be done in separate tasks 😓 These needs to be done in the backend feature itself. I am unsure why this is not the test performed in #43. This test provides more confidence and is simpler that having so many mock set up. You should read more about the test matrix to know what kind of tests to prioritize: https://nimblehq.co/compass/development/testing/#test-matrix

PR #57, opened 2023-08-08
[#54][Backend] Implement Google search page scraper

### LINKED ISSUE #54: [Backend] Implement Google search page scraper
## Why

Once the user uploads the keyword  a search result must be made. Implement this for a single keyword.

## Acceptance Criteria

- Add the scrape module that search the single keyword
- Scrape the search result and store it

## Implementation Detail
- Create a scraper with request module and parse module to scrape single keyword
- Save the result to a database

Prabeshpd:
- Close #54 

## What happened 👀

- Added the parser module to parse the raw html string
- Added the search module to request the html from google search

## Insight 📝

For a single keyword to be searched module is added for the search and parse. To search the keyword the `axios` package has been used. Interceptors has been setup for the request to have User-Agents header so that google don't block our requests. Interceptors has been used if we get any error while searching to retry 3 `(can be configured)` times exponentially backing off. Parser module has been added to parse the raw html string using cheerio

## Proof Of Work 📹
Screen Shot 2566-08-08 at 22 18 16


src/services/parser.test.ts
```diff
@@ -0,0 +1,73 @@
...
+ * @jest-environment node
+ */
+
+import { ISelectorId } from '@/types/common';
+
+import Parse from './parser';
+
+describe('Parse:', () => {
+  const selectorIds: ISelectorId = {
+    adLinkElementAttribute: '.sVXRqc',
```
tednguyendev, line 11: Mini suggestion: These selectors are repeated twice, here, and in the [config file](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/57/files#diff-615dacd7000fa06d7afef0f22b7b6d9267c6219fbfc9fa0315e51a5f5d963755R37). Should we put them all in a `constants` file? 👀

src/services/parser.test.ts
```diff
@@ -0,0 +1,73 @@
...
+    adTopParent: '#tvcap',
+    adBottomParent: '#bottomads',
+    adWords: '.uEierd',
+    nonAdWords: '.MjjYud',
+    linkElement: 'a',
+    linkElementAttribute: 'href',
+    nonAdParent: '#search',
+  };
```
olivierobert: Also minor comment to add a line break

```suggestion
  };
  
```

olivierobert (approved):
Just some minor comments but pre-approving so that you can merge once Ted approves ✅

src/services/parser.test.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/57/commits/f5bac2f13796658793eb18433cd58e9818f98a7e. I have used the one from the config file in test. As they are setup from the env file they will be one single place of truth.

src/services/parser.test.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/57/commits/b22cce6b892a7ece9198950d0dbfa77466778405

tednguyendev (approved):

PR #58, opened 2023-08-09
[#55] [Backend] Implement background job to scrape keywords asynchronously

### LINKED ISSUE #55: [Backend] Implement background job to scrape keywords asynchronously
## Why

All the keywords must be scraped and the required parsed result must be stored in database

## Acceptance Criteria
- Add job to scrape keywords asynchronously
- Make sure to store the parsed result for every keywords
- Write an integration test 
- Modify CI pipeline to integrate the database in GitHub Action

## Note
Use bull mq delayed job to prevent block of google for mass search

Prabeshpd:
- Close #55 

## What happened 👀

- Implement a job to scrape keywords
- Implement a job to add scrape job in queue for scraping multiple keywords
- Add integration test
- Modify CI pipeline to run integration test

## Insight 📝

BullMq delayed jobs is used to queue the jobs so that google does not block the mass search of keywords. For this a job for scraping is created. To add them in queue another job is created which will be used later when the file upload feature gets completed. 
For scraping an integration test is created for this twi config file for jest are created so that the mock instance of database does not block the connection with test database.
CI pipeline is modified to run the integration test that leverages postgres database

[Bull MQ delayed job](https://docs.bullmq.io/guide/jobs/delayed)

## Proof Of Work 📹

Screen Shot 2566-08-09 at 10 37 21


test/workers/scrape.test.ts
```diff
@@ -0,0 +1,41 @@
...
+    mockJob.data = { keyword: 'node', userId, id: 'a4bfbcb1-b9c5-4476-99f0-20756f4974bd' };
+    await scrape(mockJob as Job);
+
+    const data = await dbClient.searchResult.findUnique({
+      where: { id: 'a4bfbcb1-b9c5-4476-99f0-20756f4974bd' },
+    });
+
+    expect(data?.keyword).toEqual(searchResult.keyword);
+    expect(data?.html).not.toBeNull();
+    expect(data?.status).toEqual(SearchResultStatus.SUCCESS);
```
tednguyendev, line 39: Do we need the `?` here? 🤔 I think the point of doing tests is to test the assumptions we are already clear of.

test/workers/scrape.test.ts
```diff
@@ -0,0 +1,41 @@
+/**
```
tednguyendev, line 1: I think the conventional location for the `*.test.ts` files in Javascript is in the same directory of their main file(aka the `scraper.ts`), which means this file should be at `src/workers/` instead, right? 👀

test/fixtures/user.ts
```diff
@@ -0,0 +1,10 @@
+import { User } from '@prisma/client';
```
tednguyendev, line 1: So are you using both the `factories` and the `fixtures`? 🤔 I thought people only use one of them in a codebase.

src/workers/scraper.ts
```diff
@@ -1,6 +1,48 @@
...
+        nonAdUrls,
+        html: htmlData,
+      },
+    };
+
+    await updateSearchResult(args);
+  } catch (err) {
+    const args = {
+      where: {
+        userId: 1,
```
tednguyendev, line 38: This looks familiar 🙈

test/workers/scrape.test.ts
```diff
@@ -0,0 +1,41 @@
...
+
+import dbClient from '@/config/database';
+
+import { scrape } from '../../src/workers/scraper';
+import { searchResult } from '../fixtures/searchResult';
+import { user } from '../fixtures/user';
+import { SearchResultStatus } from '@/constants/enum';
+
+describe('Scrape:', () => {
+  let mockJob: Partial<Job>;
```
tednguyendev, line 15: Why should we define the `mockJob` here and instead of in the test? 👀 And why should we set it to be an empty object on the `beforeAll`? 🤔

test/workers/scrape.test.ts
Prabeshpd: It is autogenerated from the typescript. And if I remove it it complains about the type that the data might be null? So, would have to check data before every expect. So, adding it.

test/workers/scrape.test.ts
Prabeshpd: It varies according to code base and practices. For me I normally put all my test in test folder separately. Since, here I saw that most of the test in javaScript repo are under the same folder. I am keeping my unit test where the data has to be mocked under the same folder. The one with integration test I am keeping in the test directories. It also helps me to segregate the test which needs mocked database and the test that needs interactions with the actual test database.

test/fixtures/user.ts
Prabeshpd: I will remove the factories folder. 🙂

test/workers/scrape.test.ts
Prabeshpd: Since the actual type of Job in the bullMQ definition has lots of other properties and typescript complains if I do not Partially type the type of BullMq job. Since, as according to test cases I may need to add properties on the mockJob I initialized it to be empty object so can be used by others.

src/workers/scraper.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/58/commits/6cdf5cf4d079baf835a83a7591dbee5574b5c4ca

test/fixtures/user.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/58/commits/5d321ed14bc4754558d893490b6e25e2f40ab977

test/workers/scrape.test.ts
tednguyendev: 🤔 So it's another Typescript thing. Ok, thank you!

tednguyendev: 🤔 So it's another Typescript thing. Ok, thank you!

src/repositories/searchResult.test.ts
```diff
@@ -1,6 +1,6 @@
 import { createSearchResult, listSearchResult, updateSearchResult } from './searchResult';
 import { dbClientMock } from '../../test/database';
-import { searchResultFactory } from '../../test/factories/searchResult';
+import { searchResult } from '../../test/fixtures/searchResult';
```
olivierobert, line 3: It is not a blocker but the usage of factories was correct so I am unsure why moving to fixtures (with static data instead of generated data with faker)?

enerated attribute values can always be overrriden in each test. For instance, if we need a fixed ID, we can use the spread operator as done in these tests.

src/constants/enum.ts
```diff
@@ -0,0 +1,10 @@
+export enum SearchResultStatus {
+  'PENDING' = 1,
+  'SUCCESS' = 2,
+  'FAILED' = 3,
+}
+
+export enum QueueNames {
```
olivierobert, line 7: Minor comment but this enum naming is confusing. One value is for the worker queue itself as the other one is for the job. How about:

```suggestion
export enum AsyncProcess {
   'JOB' = 'scrapeJob',
   'QUEUE' = 'scraper',
}
```

olivierobert (commented):
Just some minor comment but beware of reverting to doing too many refactoring ⚠️

src/repositories/searchResult.test.ts
Prabeshpd: makes lot sense. Since, this PR has already lot of changes. Would it be Ok if I correct them on a new PR?

src/constants/enum.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/58/commits/0a59ba3636304d6fd58950850a030ae4112dd81d

src/repositories/searchResult.test.ts
olivierobert: Sure, that works 👍

olivierobert (approved):

PR #59, opened 2023-08-10
[#9][Backend] As a user, I can upload a CSV file

### LINKED ISSUE #9: [Backend] As a user, I can upload a CSV file
## Why

Once the CSV file is uploaded, the file must be validated whether the file has keywords more than 100. As the requirement is of 100 keywords the file must be validated. Once validated store the keywords in database as the user wants to view the keyword along with search result.

## Acceptance Criteria

- The uploaded keywords file must be validated for 100 keywords
- API to receive the file
- Testing must be in place

## Implementation Details

- API must use `multer` or other packages to get the file.
- Parse the file and validate 100 keywords
- Store the tags for the user in the database

### Table Columns

- id (Primary Key)
- user_id (References user table,  foreign key)
- keyword (string)

Prabeshpd:
- Close #9 

## What happened 👀

- Added an api route to upload csv file
- Added integration test with help of supertest
- Added a file lib to parse the csv file and get the array of keywords

## Insight 📝

An api route is created in which the users can upload file to parse the keywords and store them in database. For parsing the native module of `nextjs request.formData( )` is used. Since, we do not have any need to store the data somewhere and the implementation is native to framework this approach was used. A file function is used to parse the csv file to array of string of keywords.

## Proof Of Work 📹

https://github.com/nimblehq/ic-nextjs-prabesh-dahal/assets/39412015/51e0aa1f-5a20-433c-bf6b-548f6e69bb45


src/app/api/v1/upload/route.ts
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    if (tags.length > 100) {
+      const error = {
+        code: httpStatusCode.BAD_REQUEST,
+        message: 'Keywords cannot be more than 100 words',
+      };
+
+      return NextResponse.json({ error });
+    }
```
olivierobert, line 34: It would be best to keep the route/controller lean. Moving this business logic to a separate object is best. See app/actions in https://github.com/nimblehq/js-framework-benchmark/blob/develop/nextjs-13/src/app/actions/actions.tsx

src/app/api/v1/upload/route.ts
```diff
@@ -0,0 +1,48 @@
...
+    await createSearchResult(searchResultPayload);
+
+    return NextResponse.json({ message: 'File Uploaded' }, { status: httpStatusCode.OK });
+  } catch (err: any) {
+    // Todo Make a default error handler after other api are introduced
+    return NextResponse.json({
+      error: { status: httpStatusCode.INTERNAL_SERVER_ERROR, message: err.message },
+    });
+  }
+}
```
tednguyendev, line 48: I thought we were gonna use [server actions](https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions) instead of the API way 👀

src/lib/file.ts
```diff
@@ -0,0 +1,19 @@
+export async function parseCSVFormData(file: FormDataEntryValue): Promise<string[]> {
+  if (!(file instanceof Blob)) {
+    throw new Error('unable to process file');
```
tednguyendev, line 3: You can either choose to capitalize or not with the error message, but I think we should make it consistent with the [other errors](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/59/files#diff-e2edb59306703697af6f649f96e9a72a549261292404e8aac82f40ae29d2085aR9) 🙈

test/api/upload.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    await dbClient.searchResult.deleteMany({ where: { userId: userUpload.id } });
+    await dbClient.user.delete({ where: { id: userUpload.id } });
+    await dbClient.$disconnect();
+  });
+
+  beforeEach(async () => {
+    await dbClient.user.create({ data: userUpload });
+  });
+
+  it('', async () => {
```
tednguyendev, line 32: Why should we let the example name empty? 🤔

test/fixtures/user.ts
```diff
@@ -1,10 +1,19 @@
...
+export const userWorker: User = {
   id: 1,
   name: 'John',
   email: 'johndoe@gmail.com',
   password: 'password@123',
   createdAt: new Date(),
   updatedAt: new Date(),
 };
+
+export const userUpload: User = {
```
tednguyendev, line 12: Why don't we just use one `user` fixture? 👀

tednguyendev:
Btw, I think it will be better if you put the real work result in the `Proof Of Work 📹`. For example, maybe you can put an image to call the Postman? Because we already have the workflow to run the tests, putting the image to show that the tests passed is kind of, duplicated? 🙈

test/fixtures/user.ts
Prabeshpd: Jest are not running the test suite sequentially. There is an error before adding two sets of data on the database. So, using different fixtures so that the data are not duplicated or hit the constraint of table. 
I tried running the them in band to run sequentially. Although the it is run sequentially, the describe suite parts seemed to run early especially the before section where the database is cleaned and inserted with data.

Prabeshpd:
> Btw, I think it will be better if you put the real work result in the `Proof Of Work 📹`. For example, maybe you can put an image to call the Postman? Because we already have the workflow to run the tests, putting the image to show that the tests passed is kind of, duplicated? 🙈

Ok. 😄

test/fixtures/user.ts
Prabeshpd: More appropriate would be to have factories I guess as suggested by Olivier in other PR. Using factories I can create new User object using spread operator. The changed files are already too many. I am thinking of creating new chore ticket in the upcoming sprint. Would it be ok?

src/app/api/v1/upload/route.ts
tednguyendev: So you mean the server actions can't receive CSV files? 🤔

test/fixtures/user.ts
tednguyendev: I see. Thanks for your explaining ✋

src/lib/file.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/59/commits/e8b36719b56ea4f5a935b3db2aed37835eb744b0

test/api/upload.test.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/59/commits/e8b36719b56ea4f5a935b3db2aed37835eb744b0

src/app/api/v1/upload/route.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/59/commits/e8b36719b56ea4f5a935b3db2aed37835eb744b0

src/app/api/v1/upload/route.ts
tednguyendev: Understood. Thank you!

tednguyendev (approved):

src/app/actions/upload.ts
```diff
@@ -0,0 +1,37 @@
+import httpStatusCode from 'http-status-codes';
+
+import { parseCSVFormData } from '@/lib/file';
+import { createSearchResult } from '@/repositories/searchResult';
+
+export async function handleCsv(file: FormDataEntryValue | null, userId: number) {
```
olivierobert, line 6: Naming-wise, what do you think of the following:

```suggestion
export async function uploadCsv(file: FormDataEntryValue | null, userId: number) {
```

It is simple and correspond more or less to teh file name. I do not see value in having the prefix `handle` 🤔

src/app/actions/upload.ts
```diff
@@ -0,0 +1,37 @@
...
+  try {
+    if (!file) {
+      const error = { code: httpStatusCode.BAD_REQUEST, message: 'File must be sent' };
+
+      throw error;
+    }
+
+    const tags = await parseCSVFormData(file);
```
olivierobert, line 14: I am unsure why you kept wanting to name this `tags`, wouldn't the following make more sense 😅 

```suggestion
    const keywords = await parseCSVFormData(file);
```

src/lib/file.ts
```diff
@@ -0,0 +1,19 @@
...
+  if (!(file instanceof Blob)) {
+    throw new Error('Unable to process file');
+  }
+
+  const stream = file.stream() as unknown as AsyncIterable<any>;
+
+  if (file.type !== 'text/csv') {
+    throw new Error('Only Csv file is supported');
+  }
+  // Convert stream to buffer
```
olivierobert, line 11: Whenver you feel like the need to add a comment, think of extracting the code to a function:

```javascript
const convertToBuffer(stream: AsyncIterable<any>) {
   const chunks = [];

   for await (const chunk of stream) {
     chunks.push(chunk);
  }

  return Buffer.concat(chunks).toString();
}

export async function parseCSVFormData(file: FormDataEntryValue): Promise<string[]> {
  // ...
  const bufferString = convertToBuffer(stream)

  return bufferString.split(',\n').map((data) => data.trim());
}
```

It makes the code more readable 🧑‍🏫

src/lib/file.ts
```diff
@@ -0,0 +1,19 @@
+export async function parseCSVFormData(file: FormDataEntryValue): Promise<string[]> {
+  if (!(file instanceof Blob)) {
+    throw new Error('Unable to process file');
+  }
+
+  const stream = file.stream() as unknown as AsyncIterable<any>;
+
+  if (file.type !== 'text/csv') {
+    throw new Error('Only Csv file is supported');
```
olivierobert, line 9: Sinc eit is an acronym, it should be:

 ```suggestion
    throw new Error('Only CSV file is supported');
```

In addition the message could be improved to:

 ```suggestion
    throw new Error('The file is of the incorret expected type (CSV)');
```

test/actions/upload.test.ts
```diff
@@ -0,0 +1,50 @@
...
+        userId: userUpload.id,
+      },
+    });
+
+    const searchResultKeyword = data.map((datum) => ({
+      keyword: datum.keyword,
+      userId: datum.userId,
+    }));
+
+    console.log({ searchResultKeyword });
```
olivierobert, line 43: That should be removed as it looks like for debugging only

olivierobert (changes requested):

test/actions/upload.test.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/59/commits/60c2b4b73860a885a6a8b0065d51e0ec37930de6

src/lib/file.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/59/commits/60c2b4b73860a885a6a8b0065d51e0ec37930de6

src/lib/file.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/59/commits/b5f42d1ea18df0181c47a66fe2fa1549da29515f

src/app/actions/upload.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/59/commits/60c2b4b73860a885a6a8b0065d51e0ec37930de6

olivierobert (approved):

PR #60, opened 2023-08-11
[#8] [UI] As a user, I can upload csv files

### LINKED ISSUE #8: [UI] As a user, I can upload a CSV file
## Why

To run the application I would need a form to upload my CSV file so that application can get the scraped result.

## Acceptance Criteria

- Create a form Page to accept a CSV file.
- Add the UI testing

## Implementation Detail

- Create a form for uploading the CSV file.
- Use React testing library to test the form

Prabeshpd:
- Close #8 

## What happened 👀

- Added an UI component to upload files
- Added an page for upload section (Will later be app component with view of table)
- Added tests
- Added some scss (Need to prettify more will be done on viewing the results)
- Add toast notifications

## Insight 📝

We need a page to upload the csv file. So, a component to upload file is added. Fetch is used to post the file to backend. 
There will be changes on page structure and scss as more components are added with features

## Proof Of Work 📹

https://github.com/nimblehq/ic-nextjs-prabesh-dahal/assets/39412015/02066302-9181-4014-b5ee-7f65302a8181


src/app/upload/page.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+import FileUploader from '@/components/common/FileUploader';
+
+export default function Upload() {
+  const [isLoading, setIsLoading] = React.useState<boolean>(false);
+
+  const uploadFile = async (formData: FormData) => {
+    setIsLoading(true);
+    try {
+      const response = await fetch('http://localhost:3000/api/v1/upload', {
```
tednguyendev, line 14: The reason you do not create a reusable `fetch` like [this](https://github.com/nimblehq/js-framework-benchmark/blob/develop/nextjs-12/src/lib/request/manager.ts) is that you will only call API once here, since on later PRs, you will use server actions, right?

tednguyendev, line 14: Also, putting the URL in the .env file will be better.

src/app/upload/page.tsx
```diff
@@ -0,0 +1,45 @@
...
+        method: 'POST',
+        body: formData,
+        headers: { Accept: 'application/json' },
+      });
+      const data = await response.json();
+
+      toast.success(data.message, {
```
tednguyendev, line 21: As for the `toast`, we will reuse it many times. We can extract it into `/lib/toast/makeToast.ts` ✋ I learned it [from Beckham](https://github.com/nimblehq/js-framework-benchmark/pull/106#discussion_r1218939000) recently

src/components/common/FileUploader.tsx
```diff
@@ -0,0 +1,62 @@
+'use client';
+
+import * as React from 'react';
+
+import ClipLoader from 'react-spinners/ClipLoader';
+
+interface PropTypes {
+  onUpload: (formData: FormData) => void;
+  allowedMimeTypes: string[];
```
tednguyendev, line 9: Why do we need to pass the `allowedMimeTypes` around dynamically as a variable instead of hardcoding it, when we already know that the only file type we receive is csv? 🤔

src/components/common/FileUploader.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+describe('Upload', () => {
+  const onUpload = jest.fn();
+
+  it('renders component', async () => {
+    render(<FileUploader isLoading={false} onUpload={onUpload} allowedMimeTypes={['text/csv']} />);
+
+    await waitFor(() => {
+      expect(
+        screen.getByText('Upload the csv file with tags you want to search')
```
tednguyendev, line 16: We prefer the word `keywords` than `tags`, right? 🙈

src/components/common/FileUploader.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+    await onUpload(formData);
+    setFile(undefined);
+  };
+
+  return (
+    <div className="max-w-lg rounded overflow-hidden shadow-lg w-full p-4">
+      {(!isLoading && (
+        <div className="mb-4">
+          <h3 className="mb-4">Upload the csv file with tags you want to search</h3>
```
tednguyendev, line 39: Same as above, `tags` <> `keywords`

src/app/upload/page.test.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+import Upload from './page';
+
+describe('Upload', () => {
+  it('renders component', async () => {
+    render(<Upload />);
+
+    await waitFor(() => {
+      expect(
+        screen.getByText('Upload the csv file with tags you want to search')
```
tednguyendev, line 13: Same as above, `tags` <> `keywords`

src/components/common/FileUploader.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+      setFile(currentFile);
+    }
+  };
+
+  const onSubmit = async () => {
+    if (!file) return;
+    const formData = new FormData();
+    formData.append('keywords', file);
+    formData.append('userId', '1');
```
tednguyendev, line 29: hello `userId: 1` again 🙈

src/components/common/FileUploader.tsx
```diff
@@ -0,0 +1,62 @@
...
+          <input
+            type="file"
+            data-testid="file-uploader"
+            name="keywords"
+            onChange={onFileAdd}
+            accept={allowedMimeTypes[0]}
+          />
+        </div>
+      )) || <ClipLoader />}
+      <div>{file && `${file.name} - ${file.type}`}</div>
```
tednguyendev, line 49: Would it be better if we do this instead:

```jsx
file && (<div>`${file.name} - ${file.type}`</div>)
```

To avoid an empty div, I think 👀

src/components/common/FileUploader.tsx
```diff
@@ -0,0 +1,62 @@
...
+      <div className="m-1 items-center">
+        <button
+          className="bg-blue-500 hover:bg-blue-700 text-white font-bold py-2 px-4 rounded "
```
tednguyendev, line 52: Redundant space at the end ✋ 

```html
"rounded "
```

src/app/api/v1/upload/route.ts
```diff
@@ -7,7 +7,7 @@ export async function POST(request: Request) {
   try {
     const formData = await request.formData();
 
-    const file = formData.get('file');
+    const file = formData.get('keywords');
```
tednguyendev, line 10: Why do we change it from `file` to `keywords` but still use the `file` variable for it? 🤔

src/app/upload/page.tsx
Prabeshpd: Yes, I am planning to use server actions and this was one time use so I am not creating a separate lib file for it. If there is a need then I will create a separate file

src/app/upload/page.tsx
Prabeshpd: For this I was thinking the same. Since, this PR had a lot of file changes for setup. I did not make a separate file for it. For user signup there will be less file. I was planning to add it on that PR. Since, on that PR we will have a chance to use it again. 
Will that be Ok?

src/components/common/FileUploader.tsx
Prabeshpd: I wanted to make the upload component little bit dynamic supportive for future. As, it is placed under separate common component folder I wanted to make it reusable. We can remove it for the IC use case.

src/components/common/FileUploader.test.tsx
Prabeshpd: 😅 I keep getting  mixed with keywords and tags. I will add a checklist before making it ready from next time. 🙈

src/components/common/FileUploader.tsx
Prabeshpd: Here I am putting it intentionally. Since, the user login feature is not there. We are making a file directly for upload. Will keep it in a separate constant variable.

src/app/api/v1/upload/route.ts
Prabeshpd: I wanted to make a meaningful name for frontend on what we are trying to upload. In the backend I think we can rename it to `keywordFile`. I avoid using keywords for extraction because I wanted to make clear that the received object is still file and not parsed. I will change it to `keywordFile`.

src/app/upload/page.tsx
tednguyendev: Sure! Let's change it later

src/components/common/FileUploader.tsx
tednguyendev: I see. 

But in case we need to hardcode it, I think it's still better to hardcode it on the backend, since, eventually, we will get the `userId` from the session(or something familiar) on the backend, as we should not pass the `userId` from the FE and hope the BE can trust it 🙈 , like

```js
// const userId = session.user.id
const userId = '1'
```

But ya, putting it on FE is still ok for now 👍

src/app/api/v1/upload/route.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/60/commits/64bad1de85d43b7a681298c5cd1380f208f90fb3

src/app/actions/upload.ts
```diff
@@ -6,32 +6,35 @@ import { createSearchResult } from '@/repositories/searchResult';
 export async function uploadCsv(file: FormDataEntryValue | null, userId: number) {
   try {
     if (!file) {
-      const error = { code: httpStatusCode.BAD_REQUEST, message: 'File must be sent' };
+      const errorConstruct = { status: httpStatusCode.BAD_REQUEST, message: 'File must be sent' };
 
-      throw error;
+      throw errorConstruct;
```
olivierobert, line 11: What is the point of this variable renaming? `construct` is very abstract as a concert as it makes it harder to understand. IMO I do not see making such a change. This change seems uncalled for as it is not motivated by functionality to application changes 😵

src/app/page.tsx
```diff
@@ -1,6 +1,6 @@
 export default function Home() {
   return (
-    <main className="flex min-h-screen flex-col items-center justify-between p-24">
+    <main className="home">
```
olivierobert, line 3: Such classes should be added to the body class, not internal DOM elements so that whole pages can be targeted by this selector (as done by https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/60/files#diff-dc3aadb1350da3694573e4fa0ae98f4149ed4f4c586ad65d44f1db975c671275). It is not a blocker, but it is a best practice. See for a Rails project: https://github.com/nimblehq/mia-web/blob/development/engines/mia_account/app/views/layouts/mia_account/application.html.slim#L15

src/app/upload/page.test.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+import { render, screen, waitFor } from '@testing-library/react';
+
+import Upload from './page';
+
+describe('Upload', () => {
+  it('renders component', async () => {
```
olivierobert, line 8: Not correct English:

```suggestion
  it('renders the component', async () => {
```

src/components/common/FileUploader.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+    render(<FileUploader isLoading={false} onUpload={onUpload} />);
+
+    await waitFor(() => {
+      expect(
+        screen.getByText('Upload the csv file with keywords you want to search')
+      ).toBeInTheDocument();
+    });
+  });
+
+  it('calls onUpload function on submit button', async () => {
```
olivierobert, line 21: Never test implementation details ⚠️ It is better to test that a callback function is called. It seems like a minor change but if the callback prop is changed from `onUpload` to `onAfterUpload`, the test will still be current and valid.

src/components/common/FileUploader.tsx
olivierobert, line 1: Why creating a sub-directory for `common`? 🤔 Since the application is small, all components can be in the main `components` root. It is no use to over-engineer before there is a need.

src/components/common/FileUploader.tsx
olivierobert: It is better to add: `TODO: replace with current user ID` to 1) avoid raising comments 😉 and 2) ensure it is not forgotten (ost text editors can programmatically check for this).

src/components/common/FileUploader.tsx
```diff
@@ -0,0 +1,65 @@
...
+  return (
+    <div className="max-w-lg rounded overflow-hidden shadow-lg w-full p-4">
+      {(!isLoading && (
+        <div className="mb-4">
+          <h3 className="mb-4">Upload the csv file with keywords you want to search</h3>
+          <input
+            type="file"
+            data-testid="file-uploader"
+            name="keywords"
+            onChange={onFileAdd}
```
olivierobert, line 43: It is better to keep the prop value with its name:

```suggestion
            onChange={onChange}
```

src/config/ui.ts
```diff
@@ -0,0 +1,9 @@
+interface uiConfig {
+  SERVER_URL: string;
+}
+
+const config: uiConfig = {
+  SERVER_URL: process.env.REACT_BASE_ENV_URL || '',
```
olivierobert, line 6: If this is empty, then this would break. Should't an error be raised if it is missing?

olivierobert (changes requested):

src/app/page.tsx
Prabeshpd: Ok. I was also thinking the same to extract the class. Since, the changes are too much in this PR? Can I create a chore task to extract the css class to separate file. And do the styling part there?

src/components/common/FileUploader.test.tsx
Prabeshpd: Ok. Makes a lot of sense. Will avoid these kinds of test in future.

src/app/actions/upload.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/60/commits/f39f88caceb4aa855717251b62a0344199b35bca

src/components/common/FileUploader.tsx
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/60/commits/b5c65f5fc0b6b4b2cf932a7de836e723f11cbb49

src/components/common/FileUploader.test.tsx
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/60/commits/b5c65f5fc0b6b4b2cf932a7de836e723f11cbb49

src/app/page.tsx
olivierobert: Sure, that can be done in a separate chore 👍

olivierobert (approved):
Good to go now 🚀

PR #64, opened 2023-08-14
[#1] [Backend] As a user, I can sign up with email and password

### LINKED ISSUE #1: [Backend] As a user, I can sign up with email and password

Prabeshpd:
- Close #1 

## What happened 👀

- Created a server action to create user
- Added the test for user creation

## Insight 📝

For user registration we would need a method to add user for using the application. For it a server action has been created. The password are hashed and saved in database for safety reasons.

## Proof Of Work 📹

Passed CI test


test/actions/user.test.ts
```diff
@@ -0,0 +1,37 @@
...
+  });
+
+  afterEach(async () => {
+    await dbClient.user.delete({ where: { email: user.email } });
+    await dbClient.$disconnect();
+  });
+
+  it('creates the user successfully', async () => {
+    const { data: createdUser } = await addUser(userPayload);
+    const passowrdMatches = await crypt.compareWithHashValue(
```
tednguyendev, line 28: I wonder if we should use `compareWithHashValue` here.

In this test file, I think we only need to test if we successfully created a user or not, as we already have tests for the function `hash`(and should also add test for the `compareWithHashValue` in `crypt.test.ts` as mentioned above).

test/actions/user.test.ts
```diff
@@ -0,0 +1,37 @@
...
+  beforeAll(async () => {
+    userPayload = { name: user.name, email: user.email, password: user.password };
+  });
+
+  afterEach(async () => {
+    await dbClient.user.delete({ where: { email: user.email } });
+    await dbClient.$disconnect();
+  });
+
+  it('creates the user successfully', async () => {
```
tednguyendev, line 26: I think we also need to test the case when create user fails.

test/actions/user.test.ts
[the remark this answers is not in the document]
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/64/commits/18aa307b83b71eb7a4f07f1de9174e1d48936c4b

Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/64/commits/18aa307b83b71eb7a4f07f1de9174e1d48936c4b

test/actions/user.test.ts
```diff
@@ -0,0 +1,37 @@
...
+  let userPayload: Prisma.UserCreateInput;
+  beforeAll(async () => {
+    userPayload = { name: user.name, email: 'dev@nimblehq.co', password: user.password };
+  });
+
+  afterAll(async () => {
+    await dbClient.user.delete({ where: { email: 'dev@nimblehq.co' } });
+  });
+
+  it('creates the user successfully', async () => {
```
tednguyendev, line 22: How about:

```js
describe('given valid params', () => {
  it('creates the user successfully', async () => {

  });
});
```

test/actions/user.test.ts
```diff
@@ -0,0 +1,37 @@
...
+    const { data: createdUser } = await addUser(userPayload);
+
+    expect(createdUser.email).toEqual('dev@nimblehq.co');
+    expect(createdUser.name).toEqual(user.name);
+  });
+
+  it('fails to create the user with same email', async () => {
```
tednguyendev, line 29: How about:

```js
describe('given invalid params', () => {
  it('returns an internal server error', async () => {

  });
});
```

src/lib/crypt.test.ts
```diff
@@ -13,4 +13,24 @@ describe('Crypt', () => {
       expect(generatedHash).not.toBeNull();
     });
   });
+  describe('compare()', () => {
```
tednguyendev, line 16: Should it be:
```js
  describe('compare()', () => {
```

Or this 👀 : 
```js
  describe('compareWithHashValue()', () => {
```

src/lib/crypt.test.ts
```diff
@@ -13,4 +13,24 @@ describe('Crypt', () => {
...
+    it('matches the same given string with hashed string', async () => {
+      const string = 'test';
+      const expectedHashString = await hash(string);
+
+      const result = await compareWithHashValue(string, expectedHashString);
+
+      expect(result).toEqual(true);
+    });
+
+    it('does not matches a different string for the hash', async () => {
```
tednguyendev, line 26: Same as above, how about:

```js
describe('given different string', () => {
  it('does not matches', async () => {

  });
});
```

src/lib/crypt.test.ts
```diff
@@ -13,4 +13,24 @@ describe('Crypt', () => {
       expect(generatedHash).not.toBeNull();
     });
   });
+  describe('compare()', () => {
+    it('matches the same given string with hashed string', async () => {
```
tednguyendev, line 17: And:

```js
describe('given same string', () => {
  it('matches', async () => {

  });
});
```

You can change the `it('matches')` into something more detailed, also 🙈

src/app/actions/user.ts
```diff
@@ -0,0 +1,23 @@
...
+import * as userRepository from '@/repositories/user';
+
+export async function addUser(userPayload: Prisma.UserCreateInput) {
+  try {
+    const password = userPayload.password;
+    const cryptedPassword = await crypt.hash(password);
+    const payload = { ...userPayload, password: cryptedPassword };
+    const user = await userRepository.createUser(payload);
+
+    return { data: user, message: 'User is created successfully' };
```
olivierobert, line 14: How about a more user-friendly message:

```suggestion
    return { data: user, message: 'The user account has been created successfully' };
```

src/app/actions/user.ts
```diff
@@ -0,0 +1,23 @@
...
+  } catch (err: any) {
+    const error = {
+      code: httpStatusCode.INTERNAL_SERVER_ERROR,
```
olivierobert, line 17: As raised previously, never raise 500 errors. That should be left for system and unhandled errors. It is best to return 422 / unprocessable entitye.

src/lib/crypt.test.ts
```diff
@@ -13,4 +13,24 @@ describe('Crypt', () => {
       expect(generatedHash).not.toBeNull();
     });
   });
```
olivierobert, line 15: It is best to add line breaks for readability between examples:

```suggestion
  });
  
```

test/actions/user.test.ts
```diff
@@ -0,0 +1,37 @@
...
+import { addUser } from '../../src/app/actions/user';
+import { user } from '../factories/user';
+
+import dbClient from '@/config/database';
+
+describe('User:', () => {
+  let userPayload: Prisma.UserCreateInput;
+  beforeAll(async () => {
+    userPayload = { name: user.name, email: 'dev@nimblehq.co', password: user.password };
+  });
```
olivierobert, line 16: Avoid over-using `before`/`after` blocks. It is best to keep all data/fixture generation in each test. It avoids creating dependencies between tests.

test/actions/user.test.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  afterAll(async () => {
+    await dbClient.user.delete({ where: { email: 'dev@nimblehq.co' } });
```
olivierobert, line 19: How about just deleting all DB data after each test (to avoid having to put this in every test)?

test/factories/searchResult.ts
```diff
@@ -2,7 +2,7 @@ import { SearchResult } from '@prisma/client';
 
 export const searchResult: SearchResult = {
   id: 'a4bfbcb1-b9c5-4476-99f0-20756f4974bd',
-  userId: 1,
+  userId: 3,
```
olivierobert, line 5: Why does this need to be updated? Shouldn't this be created randomly and overridden in each test if necessary?

test/actions/user.test.ts
Prabeshpd: In jest when trying to run sequentially it only runs the it section sequentially. The before and after blocks may run in parallel which causes problem for other test suite if I deleted all the data.

test/actions/user.test.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/64/commits/2036f663bcba82365ed0e5fc927d09baace2f2fe

src/app/actions/user.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/64/commits/2036f663bcba82365ed0e5fc927d09baace2f2fe

src/app/actions/user.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/64/commits/2036f663bcba82365ed0e5fc927d09baace2f2fe

test/factories/searchResult.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/64/commits/8b57db944cbf1198b9f74600182b22e0bf78206a

test/actions/user.test.ts
Prabeshpd: I think it's better to have a script that populates all the required data for test. After the test has been completed then we can delete it as a teardown. At this point I think it's too much changes for it. May be we can have a separate task for it if it is a good idea

test/actions/user.test.ts
tednguyendev: I don't understand the [new change](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/commit/2036f663bcba82365ed0e5fc927d09baace2f2fe#diff-049e1a145d794c53441e1070ca61c3baeb8656b13a59aed69e91094463f871d9R12) yet.

Why should we change `describe('User:')` to `describe('given valid Params')`? If you mean that our 2 examples both happen when the params are valid, then we can change it into:

```js
describe('User:', () => {
  describe('user email is unique', () => {
    it('creates the user successfully', async () => {
    });
  });

  describe('user email is NOT unique', () => {
    it('returns with unprocessable entity error', async () => {
    });
  });
});
```

And what does [this test](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/commit/2036f663bcba82365ed0e5fc927d09baace2f2fe#diff-049e1a145d794c53441e1070ca61c3baeb8656b13a59aed69e91094463f871d9R25) means? Why the name of it is `returns with unprocessable entity error`, but we `expect` it to not have the property `error.code` === `422`? 👀 

Thank you in advance!

Prabeshpd: Oops. My bad I forgot to add the corrected test for it for checking the error.code. 🙈 Regarding the description updated the description of test in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/64/commits/53a0b88ff7d76630b9ebcb25ddf54ab1c9463f74

Prabeshpd: Yes, It's because in the previous test we have added an user. We are cleaning the database after all the test cases has passed.

tednguyendev: I see. But in testing, I don't think tests should depend on each other. I see people prefer duplicating code than trying to DRY tests and then make it couple.

I think it will be better if we clean the database after each test, not after all tests. Cleaning the database after all tests indeed helps us create fewer database records(although for an IC like this, the database creations won't be even that much), but the trade-off is that it makes the tests depend on each other, like in this case, and it is a lot harder to both the others to maintain the tests or for the code reviewers to review.

Still, if you have reasons not to clear the database after each test, I suggest you at least still create 2 users with the same email on 1 test, not on 2 tests, like this:
```js
  describe('given email is not unique', () => {
    it('returns with unprocessable entity error', async () => {
      const userPayload1 = { name: user.name, email: 'dev1@nimblehq.co', password: user.password };
      await addUser(userPayload1);
      const userPayload2 = { name: user.name, email: 'dev1@nimblehq.co', password: user.password };

      await expect(addUser(userPayload2)).rejects.toHaveProperty('code', 422);
    });
  });
```

Prabeshpd: It makes a lot sense to not to couple two testing scenarios. Thank you ted.

Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/64/commits/4d193cd0848c10680f0eacbda2771e7c976e4de6

tednguyendev (approved):

test/workers/scrape.test.ts
```diff
@@ -15,15 +15,25 @@ import { SearchResultStatus } from '@/constants/enum';
...
   let mockJob: Partial<Job>;
   let userPayload: User;
+  let searchResultPayload: SearchResult;
+  const keyword = 'node';
+  const uuid = 'a4bfbcb1-b9c5-4476-99f0-20756f4974bd';
+  const userId = 3;
 
   beforeAll(() => {
-    userPayload = { ...user, id: 1 };
+    userPayload = { ...user, id: 3 };
```
olivierobert, line 24: Shouldn't this use the constant defined above?

```suggestion
    userPayload = { ...user, id: userId };
```

This kind of complexity is why it is best not to use before block for creating fixtures. Prefer to move all of this logic to each test. There is come code repetitions but it is far easier to understand and maintain.

test/actions/upload.test.ts
```diff
@@ -14,7 +14,7 @@ describe('Upload:', () => {
   let userPayload: User;
 
   beforeAll(async () => {
-    userPayload = { ...user, id: 1 };
+    userPayload = { ...user, id: 2 };
```
olivierobert, line 17: Why does this `id` needs to be equal to 2? That is odd. This tells me that there is something wrong in the test setup.

test/actions/user.test.ts
olivierobert: I recommend fixing the test setup in this PR (or in a parallel PR that will be merged before this one) as many odd things are happening with id needing to be set to correct values 😵 

We could have something like https://flaviocopes.com/prisma-clear-database/ or https://medium.com/fuzzy-code/how-to-delete-all-the-table-contents-on-prisma-b99fff7ef7b

olivierobert (changes requested):

test/actions/user.test.ts
Prabeshpd: Got your point. It seems that I was making the test more complicated than needed. I removed all the hardcoded values to faker generated once. 

Also, for deleting the table values the describe still runs them in parallel. So, if I delete all the fields in the search_result table. It affects my test on other test suite as it  deletes all of the data even the ones that is setup or created in that test suite.

I found a different solution to this problem. Jest has introduced `globalTeardown` option. It normally takes place after the test has been completed. Have been missing on this one options. I was normally using the mocha for backend related test which does not have this problem.

test/workers/scrape.test.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/64/commits/6cc410ac2b91328f0e02761b613ac7e84945ddf1

test/actions/upload.test.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/64/commits/6cc410ac2b91328f0e02761b613ac7e84945ddf1

test/database.integration.ts
olivierobert, line 1: I recommend moving this file to `test/support/database.ts` or `test/helpers/database.ts` as technically is part of the test tooling setup, so not on the same level as other tests.

test/actions/upload.test.ts
```diff
@@ -3,38 +3,25 @@
...
-
-  afterEach(async () => {
-    await dbClient.searchResult.deleteMany({ where: { userId: userPayload.id } });
-    await dbClient.user.delete({ where: { id: userPayload.id } });
-    await dbClient.$disconnect();
-  });
-
   beforeEach(async () => {
-    await dbClient.user.create({ data: userPayload });
+    await dbClient.user.create({ data: user });
```
olivierobert, line 14: Where does the user object come from? 🤔 

Again, using before blocks causes more readability and dependency issues, so there are more downsides than benefits. One tricky part is [the order of execution](https://jestjs.io/docs/setup-teardown#order-of-execution). It is best to follow the team conventions to inline all test fixture creation into each test.

test/actions/user.test.ts
```diff
@@ -0,0 +1,27 @@
...
+import { addUser } from '../../src/app/actions/user';
+import { user } from '../factories/user';
+
+describe('User:', () => {
+  describe('given valid params', () => {
+    it('creates the user successfully', async () => {
+      const userPayload = { name: user.name, email: user.email, password: user.password };
+      const { data: createdUser } = await addUser(userPayload);
+
+      expect(createdUser.email).toEqual(user.email);
```
olivierobert, line 14: It is best to match passed values than rely on fake data when doing assertions when dealing with simple test data:

```
const userPayload = { name: user.name, email: 'dev@nimblehq.co', password: user.password };

expect(createdUser.email).toEqual('dev@nimblehq.co');
```

test/actions/user.test.ts
```diff
@@ -0,0 +1,27 @@
...
+      expect(createdUser.name).toEqual(user.name);
+    });
+  });
+
+  describe('given email is not unique', () => {
+    it('returns with unprocessable entity error', async () => {
+      const userPayload = { name: user.name, email: 'dev1@nimblehq.co', password: user.password };
```
olivierobert, line 21: Why pass a specific email if no assertion is set against it?

jest.integration.config.js
```diff
@@ -9,6 +9,7 @@ const createJestConfig = nextJest({
 // Add any custom config to be passed to Jest
 const customJestConfig = {
   setupFilesAfterEnv: ['<rootDir>/test/jest.ts'],
+  globalTeardown: '<rootDir>/test/database.integration.ts',
```
olivierobert, line 12: While that works to reset the db after the whole test suite has been executed, it does not solve the issue of deleting data between each test. As a result, some test data can still corrupt other tests. So this would need to run `afterEach` in each test file.

olivierobert (changes requested):
The test setup goes in the right direction, but there seems still to be dependencies between tests. So that must be resolved.

test/actions/user.test.ts
Prabeshpd: Here, I am testing the logic of failure in case of creating the user with same email. So, in this test I am using the same email address to create the user twice. In the second creation I get the error. I am asserting for the error code here

jest.integration.config.js
Prabeshpd: The order of execution does not make sense for the promises. As they will be handled separately. Although the order of execution runs perfectly for the things that does not have promise.

test/actions/user.test.ts
olivierobert: Understood. That makes sense then :-) A minor improvement would be to use naming to give a hint:

```suggestion
      const userPayload = { name: user.name, email: 'unique.user@nimblehq.co', password: user.password };
```

It is a small trick that make tests more readable and maintainable.

jest.integration.config.js
olivierobert: I understand and that would be fine to keep as-is as long as there are no dependencies between each test. This is the most important issue to solve.

test/database.integration.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/64/commits/956002a691762209a192135708b0bd8c1c2bbb48

test/actions/upload.test.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/64/commits/956002a691762209a192135708b0bd8c1c2bbb48

test/actions/user.test.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/64/commits/956002a691762209a192135708b0bd8c1c2bbb48

jest.integration.config.js
Prabeshpd: I have removed the dependancy between the test as you suggested. Can you have a look at it please?

test/factories/user.ts
```diff
@@ -2,7 +2,7 @@ import { faker } from '@faker-js/faker';
 import { User } from '@prisma/client';
 
 export const user: User = {
-  id: faker.number.int(),
+  id: faker.number.int(10000),
```
olivierobert, line 5: There could be still some ID collisions but with the current test suite size, it should be sufficient.

olivierobert (approved):

jest.integration.config.js
olivierobert: That looks great. Let's keep this pattern for the upcoming tests.

PR #66, opened 2023-08-15
[#6] [Backend] As a user, I can sign in with email and password

### LINKED ISSUE #6: [Backend] As a user, I can sign in with email and password
## Why

To make the application secure there must be a way to login using the credentials used for registering in the application. Only after the user is authenticated then they are allowed to perform certain tasks

## Acceptance Criteria

- Validate the username and password provided by user from database.
- After login user must be provided with JWT token
- API integration test
- Add a function to check username and password provided by user is there or not
- If there is user create a JWT token as `access_token` and `refresh_token` and pass it in response 
- Add integration tests

Prabeshpd:
- Close #6 

## What happened 👀

- Used Next-auth to create login flow
- Added test

## Insight 📝
For login the user next-auth package is used that handles the generation of JWT token by itself. It is recommended by nextJS to use this package. 
[next-auth docummentation](https://next-auth.js.org/getting-started/introduction)
## Proof Of Work 📹

CI passed test


test/services/auth.test.ts
```diff
@@ -0,0 +1,35 @@
+/**
```
Prabeshpd, line 1: add a negative scenario for user

test/services/auth.test.ts
```diff
@@ -0,0 +1,35 @@
...
+    userPayload = { name: 'Nimble Dev', email: 'nimbledev@gmail.com', password: '123456' };
+    await dbClient.user.create({ data: userPayload });
+  });
+
+  afterEach(async () => {
+    await dbClient.user.delete({ where: { email: 'nimbledev@gmail.com' } });
+    await dbClient.$disconnect();
+  });
+
+  it('returns user on given correct email and password', async () => {
```
tednguyendev, line 25: We should have tests for the unhappy cases too. Maybe we should have 3 additional tests as we are having 3 unhappy cases.

src/app/api/auth/[...nextauth]/route.ts
```diff
@@ -0,0 +1,51 @@
...
+    CredentialsProvider({
+      name: 'Credentials',
+      credentials: {
+        email: { label: 'Email', type: 'text' },
+        password: { label: 'Password', type: 'password' },
+      },
+      async authorize(credentials, _req) {
+        const response = await authenticateLogin(credentials);
+
+        if (typeof response !== 'undefined') {
```
Comment (tednguyendev) [inline src/app/api/auth/[...nextauth]/route.ts:18]: I thought in unhappy cases, we would [throw errors](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/66/files#diff-1e4285f62959d16bf21c21ca382130a7ef3675bcc812a0a100e8fd06251be758R15).

So if an unhappy case happens, can the code execute this line (18)? I thought it would stop at [line 16](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/66/files#diff-28cc05a56187d0c5aa8fa3726db269c1ac993ac57b3f2d2b413da8d4e748f66bR16) 👀 .

Comment (Prabeshpd) [reply src/app/api/auth/[...nextauth]/route.ts:18]: Yes, the code should stop at line 16. Here if there is any scenario where those cases fail and we get empty object user to add extra layer of security. 
Also the next-auth type does not let you directly process response as it may be undefined for errors. It's next-auth thing too. I believe the reason for next-auth to enforce to implement this is  the promise execution there happens in two different call stacks.

Comment (tednguyendev) [reply src/app/api/auth/[...nextauth]/route.ts:18]: Understood. Thanks for you explaination

test/services/auth.test.ts
```diff
@@ -0,0 +1,35 @@
...
+
+import dbClient from '@/config/database';
+
+describe('Auth:', () => {
+  let userPayload: Prisma.UserCreateInput;
+
+  beforeAll(async () => {
+    userPayload = { name: 'Nimble Dev', email: 'nimbledev@gmail.com', password: '123456' };
+    await dbClient.user.create({ data: userPayload });
+  });
```
olivierobert, line 18: As raised in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/64/files#r1296706821, I recommend moving this code into each test block ad revisit the teardown setup.

Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/66/commits/106d28ecbea134a48fe89af3ad683b46bf74ec44

test/services/auth.test.ts
```diff
@@ -0,0 +1,58 @@
...
+
+import { faker } from '@faker-js/faker';
+
+import { authenticateLogin } from '../../src/services/auth';
+import { user } from '../factories/user';
+import { createUser } from '../../src/app/actions/user';
+
+describe('Auth:', () => {
+  describe('given correct credentials', () => {
+    it('returns user on given correct email and password', async () => {
```
tednguyendev, line 13: The `on given correct email and password` is redundant, as we already have the `describe('given correct credentials', () => {` above

test/services/auth.test.ts
```diff
@@ -0,0 +1,58 @@
...
+      });
+
+      expect(data).not.toBeNull();
+      expect(data.email).toEqual(user.email);
+      expect(data.name).toEqual(user.name);
+    });
+  });
+
+  describe('given incorrect credentials', () => {
+    it('throws bad request error when there is no user', async () => {
```
tednguyendev, line 28: Prefer putting the conditionals(`when there is no user`) on `describe` instead of `it`, as doing that will make the tests more readable.

test/services/auth.test.ts
```diff
@@ -0,0 +1,58 @@
...
+      await expect(
+        authenticateLogin({ email: 'dev@nimblehq.co', password: 'random' })
+      ).rejects.toHaveProperty('code', 400);
+    });
+
+    it('throws bad request error if no credentials are passed', async () => {
```
tednguyendev, line 34: Same as above

test/services/auth.test.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/66/commits/abb39c4c93be09ded22c21b475602858a4c63ab6

test/services/auth.test.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/66/commits/abb39c4c93be09ded22c21b475602858a4c63ab6

tednguyendev: As we are more used to use [the word `given`](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/66/files#diff-c8e8e91247228cd72688e9c49c884efd98e04cbed2d5e7f901060bbd05729605R12) for a `describe`, let's add the final change is to change from [`when`](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/66/files#diff-c8e8e91247228cd72688e9c49c884efd98e04cbed2d5e7f901060bbd05729605R27) to `given` ✋

tednguyendev (approved):

test/services/auth.test.ts
```diff
@@ -0,0 +1,62 @@
...
+import { faker } from '@faker-js/faker';
+
+import { authenticateLogin } from '../../src/services/auth';
+import { user } from '../factories/user';
+import { createUser } from '../../src/app/actions/user';
+
+describe('Auth:', () => {
+  describe('given correct credentials', () => {
+    it('returns user', async () => {
+      const userPayload = { name: user.name, email: user.email, password: user.password };
```
olivierobert, line 14: Add a line break between each block to delimit each part of [the AAA structure of tests](https://medium.com/@pjbgf/title-testing-code-ocd-and-the-aaa-pattern-df453975ab80).

```suggestion
      const userPayload = { name: user.name, email: user.email, password: user.password };
      
```

test/services/auth.test.ts
```diff
@@ -0,0 +1,62 @@
...
+      await expect(authenticateLogin({ email: '', password: '' })).rejects.toHaveProperty(
+        'code',
+        400
+      );
+    });
+  });
+
+  describe('when passed password is incorrect', () => {
+    it('throws bad request error', async () => {
+      const email = faker.internet.email();
```
olivierobert, line 46: Is that needed since faker will generate an email anyway?

test/services/auth.test.ts
```diff
@@ -0,0 +1,62 @@
...
+      const userPayload = {
+        name: user.name,
+        email,
```
olivierobert, line 49: I recommend to set it to a fixed email address:

```suggestion
        email: 'user@nimblehq.co',
```

test/services/auth.test.ts
```diff
@@ -0,0 +1,62 @@
...
+        password: user.password,
+      };
+      await createUser(userPayload);
+
+      await expect(
+        authenticateLogin({
+          email: email,
```
olivierobert, line 56: I recommend to set it to a fixed email address:

```suggestion
          email: 'other.user@nimblehq.co',
```

olivierobert (commented):
Some minor issues in tests. This still required fixing the test setup in #64

test/services/auth.test.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/66/commits/60c2f10bae5e1863561ada4099a6b2bacb68772b

test/services/auth.test.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/66/commits/60c2f10bae5e1863561ada4099a6b2bacb68772b

olivierobert (approved):

PR #70, opened 2023-08-18
Release - 0.4.0

Prabeshpd:
https://github.com/nimblehq/ic-nextjs-prabesh-dahal/milestone/4

## Features
- #60 
- #64 
- #65 
- #66 
- #67 

## Chores
- #63 

## Bugs

`N/A`

PR #72, opened 2023-08-21
 [#70] [Chore] Setup test environment with database schema per test suite

### LINKED ISSUE #71: [Chore] Setup test environment with database schema per test suite
## Why

 There are still some flaky scenarios if someone wanted to use any unique constraint then they had to make sure those values would not match on each suite.

## Acceptance Criteria
Create schema under same database for each test suite.

## Resources

[example]( https://ericyd.hashnode.dev/testing-against-a-database-new-approaches)

Prabeshpd:
- Close #71 

## What happened 👀

- Created mock for prisma connect.
- Passed the environment variable for each schema creation with unique schema ID
- Delete the schema and data after each test

## Insight 📝

Previously we used to make sure that any unique constraints are not repeated entire test suits. Now, each test suite have their own schema which gets deleted after the test is over. [Integration Test Reference](https://blog.ludicroushq.com/a-better-way-to-run-integration-tests-with-prisma-and-postgresql)

## Proof Of Work 📹
Passed CI test


tednguyendev (approved):

src/config/__mocks__/database.ts
```diff
@@ -0,0 +1,40 @@
+import { execSync } from 'child_process';
+import { join } from 'path';
+
+import { PrismaClient } from '@prisma/client';
+import { v4 } from 'uuid';
+
+const generateDatabaseURL = (schema: string) => {
+  if (!process.env.POSTGRES_DB_CONNECTION_URL) {
+    throw new Error('please provide a database url');
```
olivierobert, line 9: Minor Developer Experience improvemebt suggestion (oen should have clear error messages):

```suggestion
    throw new Error('`POSTGRES_DB_CONNECTION_URL` has not been defined');
```

src/config/__mocks__/database.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+  const url = process.env.POSTGRES_DB_CONNECTION_URL + `?schema=${schema}`;
+
+  return url.toString();
+};
+
+const schemaId = `test-${v4()}`;
+const prismaBinary = join(__dirname, '..', '..', '..', 'node_modules', '.bin', 'prisma');
+const url = generateDatabaseURL(schemaId);
+process.env[' POSTGRES_DB_CONNECTION_URL'] = url;
```
olivierobert, line 19: 1. Is the space intentional?

```suggestion
process.env['POSTGRES_DB_CONNECTION_URL'] = url;
```

2. So this will override the locally defined env variable? If yes, it would be good to add a comment here to let future devs know what is going and to add more emphasis on a potntial breaking bit:
```suggestion

# Override locally defined environment variable
process.env[' POSTGRES_DB_CONNECTION_URL'] = url;
```

Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/72/commits/dd79cd103a872db6a10e4e4e104e624980c6d91a

src/config/__mocks__/database.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/72/commits/dd79cd103a872db6a10e4e4e104e624980c6d91a

olivierobert (approved):

PR #75, opened 2023-08-22
[#13] [Backend] As a user, I can view the list of search results

### LINKED ISSUE #13: [Backend] As a user, I can view the list of search results
## Why
Once the csv file is uploaded and saved user may want to see the keywords and search result
## Acceptance Criteria

- Server actions to view search_result
- Pagination must be implemented
- Search with tags must be implemented
- Test must be there

Prabeshpd:
- Close #13 

## What happened 👀

- Created a pagination module
- Created a buildQuery module to build query for search result
- Created  a server action to list search result with features of pagination and search Filter.
- Add Tests.

## Insight 📝

Created a server actions to fetch the search results. To fetch the search result there is a need to incorporate the pagination and search feature on keyword as the data may be too big to view for users. To build the query a module has been defined inside of the query folder. This was done as mentioned in compass and followed other repository of rails. 

## Proof Of Work 📹

Passed CI pipelines


test/services/searchResultEntity.test.ts
```diff
@@ -0,0 +1,91 @@
...
+import { createSearchResult } from '@/repositories/searchResult';
+
+import { userFabricator } from '../factories/user';
+import { searchResultFabricator } from '../factories/searchResult';
+
+describe('listSearch', () => {
+  describe('given valid filter pagination params and user id', () => {
+    it('returns the valid data after filter', async () => {
+      const user = userFabricator();
+      const createdUser = await createUser(user);
```
tednguyendev, line 15: Why do you need to create the `userFabricator`?

Even when we need to create a user with a custom email, I think we can reuse the existing `user` factory:

```js
const createdUser = await createUser({ ...user, email: 'abc@gmail.com' });
```

And we don't need to create a user with a custom email here, so using `user` may be more suitable.

src/app/actions/searchResult.ts
```diff
@@ -0,0 +1,20 @@
+import withAuth from '@/lib/handler/auth.action';
+import { listSearch } from '@/services/searchResultEntity';
+import { QuerySearchWithPaginationParams } from '@/types/query';
+
+export async function listResults(
```
tednguyendev, line 5: Why don't you name the function name(`listResults`) and the file name(`searchResult`) the same for more consistency? 👀 

Is that a JS convention? 🤔

For the [upload action](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/blob/develop/src/app/actions/upload.ts), I saw you named the file and the function name differently too(`upload.ts` + function `uploadCsv`)

src/queries/searchResultQuery.ts
```diff
@@ -0,0 +1,31 @@
+class SearchResultQuery {
```
tednguyendev, line 1: Correct me if I'm wrong, but the purpose of this whole class is just to create a hash, right? 🙈 

How about using something simpler, like, a function?

```js
function getQuery(params) {
  const query = {}

  if (params.userId) {
    query['userId'] = params.userId;
  }

  if (params.keyword) {
    query['keyword'] = { contains: params.keyword };
  }

  return query
}
```

src/services/searchResultEntity.ts
```diff
@@ -0,0 +1,40 @@
...
+) {
+  try {
+    const { currentPage = 1 } = querySearchWithPaginationParams;
+    const paginationQueryParams = getPaginationParams(querySearchWithPaginationParams);
+    const authenticatedQueryFilterParams = {
+      ...querySearchWithPaginationParams.queryParams,
+      userId,
+    };
+
+    const searchResultQuery = new SearchResultQuery(authenticatedQueryFilterParams);
```
tednguyendev, line 22: Continue with the comment above. And instead of using this class as:

```js
const searchResultQuery = new SearchResultQuery(authenticatedQueryFilterParams);
const query = searchResultQuery.construct();
```

We can just:
```js
const searchResultQuery = getQuery(authenticatedQueryFilterParams);
```

src/services/searchResultEntity.ts
```diff
@@ -0,0 +1,40 @@
...
+import { QuerySearchWithPaginationParams } from '@/types/query';
+
+export async function listSearch(
+  userId: number,
+  querySearchWithPaginationParams: QuerySearchWithPaginationParams<{
+    keyword?: string;
+  }>
+) {
+  try {
+    const { currentPage = 1 } = querySearchWithPaginationParams;
```
tednguyendev, line 15: This snippet is a bit painful to read 🤣

The variable names are just so long:
- `authenticatedQueryFilterParams`
- `getPaginationParams(querySearchWithPaginationParams)` 

Remind me of Java's method names 🙈

src/app/actions/searchResult.ts
```diff
@@ -0,0 +1,20 @@
...
+  querySearchWithPaginationParams: QuerySearchWithPaginationParams<{
```
tednguyendev, line 6: Why do we call it `querySearchWithPaginationParams` instead of something like, `params`? 👀 Is that a convention?

src/repositories/searchResult.test.ts
```diff
@@ -27,30 +29,43 @@ describe('Search Result Respository', () => {
...
+    it('returns total count as zero with no matching id', async () => {
+      const userId = 1;
+
+      dbClientMock.searchResult.count.mockResolvedValue(0);
+
+      await expect(countSearchResult({ userId })).resolves.toEqual(0);
+    });
+  });
+
+  describe('update Search Result', () => {
```
tednguyendev, line 66:
```suggestion
  describe('Update Search Result', () => {
```

tednguyendev: Plus, this [`describe('createSearchResult', () => {`](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/blob/develop/src/repositories/searchResult.test.ts#L6)

Should be fixed too: 

```js
  describe('Create Search Result', () => {
```

test/services/searchResultEntity.test.ts
```diff
@@ -0,0 +1,91 @@
...
+ */
+import { listSearch } from '@/services/searchResultEntity';
+import { createUser } from '@/repositories/user';
+import { createSearchResult } from '@/repositories/searchResult';
+
+import { userFabricator } from '../factories/user';
+import { searchResultFabricator } from '../factories/searchResult';
+
+describe('listSearch', () => {
+  describe('given valid filter pagination params and user id', () => {
```
tednguyendev, line 12:
```suggestion
  describe('given valid filter pagination params and userId', () => {
```

test/services/searchResultEntity.test.ts
```diff
@@ -0,0 +1,91 @@
...
+ * @jest-environment node
+ */
+import { listSearch } from '@/services/searchResultEntity';
+import { createUser } from '@/repositories/user';
+import { createSearchResult } from '@/repositories/searchResult';
+
+import { userFabricator } from '../factories/user';
+import { searchResultFabricator } from '../factories/searchResult';
+
+describe('listSearch', () => {
```
tednguyendev, line 11: The structure of your test has room for improvement 👀 

Instead of:

```js
describe('given valid filter pagination params and user id', () => {
  it('returns the valid data after filter', async () => {
  });
});

describe('given userId with no search results associated with user', () => {
  it('returns empty data', async () => {
  });
});

describe('given only pagination params', () => {
  it('returns all the data with pagination', async () => {
  });
});
```

How about:

```js
describe('given ONLY paginate', () => {
  it('returns correct response', async () => {
  });
});

describe('given paginate AND search', () => {
  describe('given user have search results', () => {
    it('returns correct response', async () => {
    });
  });

  describe('given user NOT have search results', () => {
    it('returns correct response', async () => {
    });
  });
});
```

src/queries/searchResultQuery.ts
tednguyendev, line 1: Same as above, we are having tests for lib files.

src/app/actions/searchResult.ts
Prabeshpd: No actually I want to have all the functions of searchResult entity or table here. so, that's why the file name and function name are different. If it was the default function of file would have the same function name as that of file.

src/queries/searchResultQuery.ts
Prabeshpd: Yes, In future there will be other parameters too. So, ended up creating class. In the future the function would be too big with many if conditions. Just wanted to avoid it.

src/app/actions/searchResult.ts
Prabeshpd: Just wanted to make it more meaningful with the contents that it has. Seems I have created quite a long name. Will try to come up with meaningful short name or replace it with `query`.

src/queries/searchResultQuery.ts
tednguyendev: I wonder how many more params will we have that we need to use a class 🤔

But yes, that will work, too ✋ It's just that I prefer keeping things simple. And for a function with the purpose is to just transforming params, normally, I will try to delay making it a class as long as I can 😄

src/repositories/searchResult.ts
```diff
@@ -18,13 +20,18 @@ export const createSearchResult = async (
...
+    skip: paginationParams.offset,
+    take: paginationParams.limit,
   });
 
-  return data;
-};
+export const countSearchResult = async (query: { [key: string]: any } = {}) =>
+  dbClient.searchResult.count({ where: query });
+
+export const listAllPendingSearchResult = async (userId: number) =>
```
olivierobert, line 36: How about making it more generic (so that it can handle any status and setting a default):

```suggestion
export const listSearchResultByStatus = async (userId: number, status: string = SearchResultStatus.PENDING) =>
```

src/types/query.ts
olivierobert, line 1: SHouldn't this be called `pagination.ts` instead as it mostly related to that topic 🤔 ?

src/repositories/searchResult.ts
```diff
@@ -18,13 +20,18 @@ export const createSearchResult = async (
...
-    },
-    select: searchResultScraping,
+export const listSearchResult = async (
+  query: { [key: string]: any } = {},
+  paginationParams: PaginationQueryParams
+) =>
+  dbClient.searchResult.findMany({
+    where: query,
+    skip: paginationParams.offset,
+    take: paginationParams.limit,
```
olivierobert, line 30: How about having default values so that we don't need to pass them all the time? The defaults should be set as constant or env variables.

test/factories/searchResult.ts
```diff
@@ -16,3 +16,19 @@ export const searchResult: SearchResult = {
   nonAdUrls: [],
   status: 1,
 };
+
+export const searchResultFabricator = (userId?: number, keyword?: string) => ({
```
olivierobert, line 20: Why is a new factory needed? The existing one can be used

src/queries/searchResultQuery.ts
```diff
@@ -0,0 +1,31 @@
...
+  queryParams: any;
+  searchResultQuery: any;
+
+  constructor(queryParams: any = {}) {
+    this.queryParams = queryParams;
+    this.searchResultQuery = {};
+  }
+
+  construct() {
```
olivierobert, line 10: That is an odd name since it is very close to `constructor` 💣  How about simply:

```suggestion
  build() {
```

src/queries/searchResultQuery.ts
olivierobert, line 1: That object does not do any querying 😅 It is more like a helper in its current state. I am unsure if it has more value than just using the exiting repository object 🤔

src/queries/searchResultQuery.ts
```diff
@@ -0,0 +1,31 @@
...
+    this.filterByUserId();
+    this.filterByKeywords();
+
+    return this.searchResultQuery;
+  }
+
+  filterByUserId() {
+    if (this.queryParams.userId) {
```
olivierobert, line 18: Prefer the early return syntax:

```javascript
filterByUserId() {
  const userId = this.queryParams.userId;

  if (!tuserId) return;

  this.searchResultQuery['userId'] = userId;
}
```

src/queries/searchResultQuery.ts
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  construct() {
+    this.filterByUserId();
+    this.filterByKeywords();
+
+    return this.searchResultQuery;
+  }
+
+  filterByUserId() {
```
olivierobert, line 17: How about making these classes private:

```suggestion
  #filterByUserId() {
```

Or just using the standard prefix `_` to explicitly tell that these methods should not be accessed:

```suggestion
  _filterByUserId() {
```

olivierobert (changes requested):
Many implementation issues so you will need to revise the implementation entirely

src/services/searchResultEntity.ts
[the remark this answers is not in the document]
Prabeshpd: I am a little confused on this one. I understood your point on making a query object and creating the searchResultQuery from it. So, as I understand I can create searchQuery object and call the repository to fetch the data from the actions page handler itself right? 
The problem is same piece of code is utilised in API as well. I wanted to disintegrate the logic of formatting data or raising some custom errors for end response or result. Normally I used to put this in service layer in my previous projects. If I want to decouple the logic then for such scenarios where should I put them instead of service layer? Can you guide me on this part please?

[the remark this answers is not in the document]
olivierobert: As its name says, a query object is meant to abstract complex DB querying logic. This is usually used in controllers to keep them lean. However, in the current implementation, the query object does not do any DB query but is merely a helper.

As per [my comment raised below](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/75#discussion_r1302459448), an option is to use the existing repository: add params for pagination with default values and perform the DB query. The output should be a list of records, so not structured as an API response:

```
data: searchResults,
meta: { totalCounts: totalSearchResults, currentPage, limit: paginationQueryParams.limit },
```

This response structure is only needed for the API response. On a regular Web controller, we would need two props: 1) search results and 1) pagination (current page, total pages, current page size) Treats this two props as distinct items.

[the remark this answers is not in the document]
Prabeshpd: Okay. Got the point. Thank you. Will correct the PR.

src/repositories/searchResult.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/75/commits/8dbebf789e17df5f983c903fad85b20f56d1c482

src/repositories/searchResult.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/75/commits/8dbebf789e17df5f983c903fad85b20f56d1c482

src/services/searchResultEntity.ts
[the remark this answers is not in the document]
Prabeshpd: I have removed the function from the service layer. Apart from that I have disintegrated the web controller to two distinct actions to get the count and results. I have transformed the helper query builder to function and moved it to lib folder. I believe it is much simpler this way. If the needs arises then we can convert this to query object and use it to query the database. For this use case I think this is extra work. Can you have a look at it please? cc @olivierobert , @tednguyendev

test/repository/searchResult.test.ts
```diff
@@ -0,0 +1,152 @@
...
+        const customQuery = getSearchResultQuery({ userId: createdUser.id, keyword: 'no' });
+        const paginationQuery = getPaginationParams({ maxRows: 3 });
+
+        const data = await listSearchResult({ customQuery, paginationQuery });
+
+        expect(data.length).toEqual(3);
+      });
+    });
+
+    describe('given user have no search Result', () => {
```
tednguyendev, line 106:
```suggestion
    describe('given user have no search results', () => {
```

As you are already using the world "results" instead of "Result" [above](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/75/files#diff-ecc7ec255649de569928eaa447a954face611cedeaf5d9f51bd987b921e5b5deR59).

test/repository/searchResult.test.ts
```diff
@@ -0,0 +1,152 @@
...
+import { createSearchResult, listSearchResult } from '@/repositories/searchResult';
+
+import { user } from '../factories/user';
+import { searchResult } from '../factories/searchResult';
+import { getSearchResultQuery } from '@/lib/searchResultQuery';
+import { getPaginationParams } from '@/lib/pagination';
+
+describe('list Search Result', () => {
+  describe('given only pagination params', () => {
+    it('returns all the data with pagination.', async () => {
```
tednguyendev, line 14:
```suggestion
    it('returns all the data with pagination', async () => {
```

(Remove the dot)

src/lib/searchResultQuery.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  if (params.userId) {
+    query['userId'] = params.userId;
+  }
+
+  if (params.keyword) {
+    query['keyword'] = { contains: params.keyword };
+  }
+
+  return query;
```
tednguyendev, line 22: Looks a lot neater! 😄

src/lib/pagination.test.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * @jest-environment node
+ */
+
+import { getPaginationParams } from './pagination';
+
+describe('Pagination:', () => {
+  describe('given no pagination params', () => {
+    it('returns default pagination query parameters', () => {
```
tednguyendev, line 9: The test structure is improving 💪

tednguyendev (approved):
Just some minors.

This PR is a lot more readable now 👍

tednguyendev (approved):
LGTM now 👍

test/repository/searchResult.test.ts
olivierobert, line 1: Why are these tests not in `repositories/searchResult.test.ts`? I would be best to have all tests for the same file together.

olivierobert (changes requested):

src/lib/searchResultQuery.test.ts
```diff
@@ -0,0 +1,39 @@
...
+
+      expect(query).toEqual({});
+    });
+  });
+
+  describe('given only userId as parameter', () => {
+    it('returns query containing userId', () => {
+      const query = getSearchResultQuery({ userId: 1 });
+
+      expect(query).toEqual({ userId: 1 });
```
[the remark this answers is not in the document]
Prabeshpd: It doesnot actually query the database. It is a helper. It is only testing whether the resulting object from helper function is in place. I think it will be much clearer if I assert the resulting query does not have userId with value 2 or other prperties.

test/repository/searchResult.test.ts
Prabeshpd: These are unit tests. These test are mocking the database implementation. The other are integration test that interact with database and need separate database instance. I personally prefer the integration testing for such scenarios. I see here the mocking is done and tested in the js-framework that's why kept them.

src/lib/searchResultQuery.test.ts
```diff
@@ -0,0 +1,39 @@
...
+  describe('given no query parameters', () => {
+    it('returns empty object', () => {
+      const query = getSearchResultQuery();
+
+      expect(query).toEqual({});
+    });
+  });
+
+  describe('given only userId as parameter', () => {
+    it('returns query containing userId', () => {
```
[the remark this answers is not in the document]
Prabeshpd: List search result would actually return the data after pagination. So, it does not have information on exactly how much data is present in the database. Should I return the count from the action function as well. Since, last time you suggest to treat them as two separate props for presentation layer. I used separate function. I can merge them to one action. U think that makes way more sense.

src/services/searchResultEntity.ts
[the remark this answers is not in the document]
Prabeshpd: List search result would actually return the data after pagination. So, it does not have information on exactly how much data is present in the database. Should I return the count from the action function as well. Since, last time you suggest to treat them as two separate props for presentation layer. I used separate function. I can merge them to one action. U think that makes way more sense.

src/lib/searchResultQuery.test.ts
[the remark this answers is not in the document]
Prabeshpd: It is only a helper function to return the query. 😅. I think right now it will be complicated to make the query object to query the database. In the further filter process when there is need I plan to create the query object to filter lists.

src/services/searchResultEntity.ts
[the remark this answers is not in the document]
olivierobert: Let's clarify why we need a total count of search results  😅  What is the purpose? It seems to be for setting meta data (https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/81/files#diff-082b2577a87e52b1266cce36139a4a71c7275509b6e8937ee29d9e93e35f22dcR46) but I am unsure how it is used 🤔 

What I mean is that it seems redundant to have a server action to do a count as one could use the existing server action (without any pagination) and do a count on the client side.

src/lib/searchResultQuery.test.ts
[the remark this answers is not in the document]
olivierobert: Indeed. I get it now :-)

test/repository/searchResult.test.ts
olivierobert: The team does not draw a difference between unit and integration testing. All of them are considered non-UI tests. I recommend you to reach https://nimblehq.co/compass/development/testing/#test-pyramid and check this presentation: https://docs.google.com/presentation/d/1cM_VHVYHSw7WLMy9XRpqtWc5NIkBlifdcAyJ4gCioVg/edit?usp=sharing

So there is no need to separate these tests to a separate file. If a DB must be used for specific tests to have more confidence, if no don't.

src/lib/searchResultQuery.test.ts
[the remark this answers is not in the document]
olivierobert: Same concern as above so all is good

test/repository/searchResult.test.ts
Prabeshpd: Ok. makes more sense. the test with mocked database don't add much value. Will change those to use database instance. I think it's better to do those in separate task.

src/services/searchResultEntity.ts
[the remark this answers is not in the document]
Prabeshpd: To handle the pagination we would need to know how many data are present totally in database. So, the fetch of results first will give the user 10 sets of data. So, the count the frontend has is total of 10 length. But in database there are lots of data. so, to know how many pages of data are there we would need the count of total data based on filters excluding pagination. So, I would normally get the query and return from the same function that is returning the total count of data. 😅  For, now I will be returning them from the one single action and update the PR.

test/repository/searchResult.test.ts
Prabeshpd: I really liked the presentation. 😄  I saw lot of mockings in js-framework repo so followed that one. Will update my few test cases to add more value. Thank you

src/constants/constants.ts
```diff
@@ -0,0 +1,3 @@
+export const PAGE_LIMIT = 10;
+export const DEFAULT_OFFSET_ROWS = 0;
+export const DEFAULT_CURRENT_PAGE = 1;
```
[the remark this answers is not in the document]
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/75/commits/77ad23529ead354c06bec249a7d91a3de103567d

src/lib/searchResultQuery.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+
+export const getSearchResultQuery = (params: QueryParams = {}) => {
+  const query: Query = {};
+
+  if (params.userId) {
+    query['userId'] = params.userId;
+  }
+
+  if (params.keyword) {
```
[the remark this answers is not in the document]
Prabeshpd: Ok.

src/services/searchResultEntity.ts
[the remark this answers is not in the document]
Prabeshpd: I have updated my code. Can you have a look at it please?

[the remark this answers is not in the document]
olivierobert: I fully understood the concern about having paginated results not being compatible with the need to have a total count. That is not the issue. What I mean it is redundant to have two fetch methods. We could count the total results by just using `listSearchResult`. 

All attributes are optional in `searchResultQueryListParams`:

```typescript
export interface searchResultQueryListParams {
  customQuery?: { [key: string]: any };
  paginationQuery?: PaginationQueryParams;
}
```

So `listSearchResult` could recent no `paginationQuery` if necessary (which is the same as what `countSearchResult` operates).

To make that happen `listSearchResult` could need to be modified:

```typescript
export const listSearchResult = async (query: searchResultQueryListParams = {}, isPaginated:bool = true) => {
  const { customQuery, paginationQuery } = query;

  return dbClient.searchResult.findMany({
    where: customQuery,
    skip: paginationQuery?.offset || PAGINATION_OFFSET_ROWS,
    take: paginationQuery?.limit || PAGINATION_LIMIT,
  });
};
```

So the code could be:

```typescript
const searchResults = await listSearchResult({
  customQuery,
  paginationQuery,
});
const count = await listSearchResult(customQuery, false).length
```

The current approach works but I wanted to raise a different alternative 💡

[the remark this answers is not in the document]
Prabeshpd: Ok. Understood your point. Just a little concern here. Doesn't it make the function to look a little weird and unreadable to add boolean as the argument. I generally tend to avoid these arguments. Just wanted to know are these good patterns to have or not?

test/repository/searchResult.test.ts
Prabeshpd: For those tests there are two separate configs. If I move these tests then I would need to add the setup of data related to database. As the current test structure as unit config and integration config. The unit config mocks the prisma implementation completely and they look into folder for which configuration to use. If I move them to test folder they will follow the test with real database. Will it be okay if I did in the new pR creating new task?

src/services/searchResultEntity.ts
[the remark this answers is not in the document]
olivierobert: By weirdness, (I assume that) you mean that it could be hard to understand what this option is for without looking at the implementation. That could be improved in several ways:

- Passing an object for options:

```
export interface searchQueryListOption {
  isPaginated: bool;
}

const defaultSearchQueryListOption: searchQueryListOption = {
    isPaginated: true,
};

export const listSearchResult = async (query: searchResultQueryListParams = {}, option = defaultSearchQueryListOption) => {
  const { customQuery, paginationQuery } = query;

 //...
};
```

To me this is less weird that duplicating the code to fetch all search details 😅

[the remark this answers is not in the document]
Prabeshpd: I have implemented these changes. I tried decorator but I think it is too complex for now. So, moved to solution of yours. 😅 . I got you point and it makes a lot sense too. Thank you. Can you have a look at it now?

src/repositories/searchResult.ts
```diff
@@ -18,13 +34,27 @@ export const createSearchResult = async (
...
-  return data;
+  if (options.isPaginated) {
+    return dbClient.searchResult.findMany({
+      where: customQuery,
+      skip: paginationQuery?.offset || PAGINATION_OFFSET_ROWS,
+      take: paginationQuery?.limit || PAGINATION_LIMIT,
+    });
+  }
+
+  return dbClient.searchResult.findMany({ where: customQuery });
```
olivierobert, line 51: Minor sugegstion: repetitions for `{ where: customQuery }` can be avoided with:

```javascript
let findQuery = { where: customQuery };

if (options.isPaginated) {
  findQuery = {...findQuery, {
      skip: paginationQuery?.offset || PAGINATION_OFFSET_ROWS,
      take: paginationQuery?.limit || PAGINATION_LIMIT,
  }};
}

return dbClient.searchResult.findMany(findQuery);
```

olivierobert (commented):
Just a minor suggestion to consider. It is not compulsory so up to you to decide.

olivierobert (approved):

PR #77, opened 2023-08-22
[#76] [Backend] Set up auth for server actions

### LINKED ISSUE #76: [Backend] Setup auth for server actions
## Why

User must be authenticated to use features such as uploading the files or listing the search results. Since, we are using server actions for most of the interaction of frontend and backend. We need a mechanism to authenticate server actions

## Acceptance Criteria

- Add authentication function for server action using next-auth
- Add test

Prabeshpd:
- Close #76 

## What happened 👀

- Add authentication function for server actions.
- Add tests

## Insight 📝

As the user needs to be authenticated a authentication function is also added under the `lib/handler` folder. Next auth is used to authenticate the users. 

## Proof Of Work 📹

Passed CI pipelines


src/repositories/user.test.ts
```diff
@@ -18,4 +18,23 @@ describe('User Respository', () => {
       await expect(createUser(userAttributes)).resolves.toEqual(userPayload);
     });
   });
+
+  describe('findUser', () => {
+    it('finds User with matching id', async () => {
+      const id = 1;
+
+      const userAttributes = {
```
tednguyendev, line 26: The `userAttributes` only need the id, right? So how about

```js
const userAttributes = {
  id: 1,
};
```

tednguyendev (approved):
LGTM 👍

olivierobert (commented):
Just one minor issue and this can be merged

src/repositories/user.test.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/77/commits/417862b9df3f688746db755c74724b516f290812

PR #78, opened 2023-08-22
[#73] [Backend] As a user, I can view the detail of single keyword

### LINKED ISSUE #73: [Backend] As a user, I can view the detail of single keyword
## Why

User need to view the scraped result of single keyword.

## Implementation Details

- Create a server function to fetch detail of keyword.
- Write test for action.

Prabeshpd:
- Close #73 

## What happened 👀

-  Create a feature to fetch detail of keyword
-  Add test

## Insight 📝

Add a function to fetch the detail of keyword. The function is authenticated and if the userId of the search result does not match with the current user Id an `Unauthorised` error is thrown. 

## Proof Of Work 📹

Passed CI pipelines test


test/services/searchResultEntity.test.ts
```diff
@@ -89,3 +92,50 @@ describe('listSearch', () => {
...
 });
+
+describe('findById:', () => {
+  describe('given valid id and user id associated with result', () => {
+    it('returns valid data', async () => {
+      const user = userFabricator();
+      const createdUser = await createUser(user);
+
+      const searchResultPayload = searchResultFabricator(createdUser.id, 'node');
+      const createdSearchResult = await createSearchResult(searchResultPayload);
```
tednguyendev, line 103: What is the purpose of the function `createSearchResult`? Why did we create it in this PR? 🤔

test/services/searchResultEntity.test.ts
```diff
@@ -89,3 +92,50 @@ describe('listSearch', () => {
     });
   });
 });
+
+describe('findById:', () => {
+  describe('given valid id and user id associated with result', () => {
```
tednguyendev, line 97: This structure can also be benefited from the reorganization like in [the previous PR](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/75#discussion_r1301393236) 👀

src/services/searchResultEntity.ts
```diff
@@ -38,3 +41,41 @@ export async function listSearch(
...
+        code: httpStatusCode.UNAUTHORIZED,
+        message: 'Unauthorized to view the keyword detail.',
+      };
+
+      throw error;
+    }
+
+    return searchResult;
+  } catch (err: any) {
+    console.log(err);
```
tednguyendev, line 69: Friendly reminder to remove the loggings 👍

src/repositories/searchResult.ts
```diff
@@ -35,3 +38,5 @@ export const countSearchResult = async (query: { [key: string]: any } = {}) =>
 
 export const listAllPendingSearchResult = async (userId: number) =>
   dbClient.searchResult.findMany({ where: { userId, status: SearchResultStatus.PENDING } });
+
+export const findById = async (id: string) => dbClient.searchResult.findUnique({ where: { id } });
```
tednguyendev, line 42: We should also have test for this new function ✋

test/services/searchResultEntity.test.ts
```diff
@@ -89,3 +92,50 @@ describe('listSearch', () => {
...
+
+  describe('given invalid id', () => {
+    it('rejects with not found error', async () => {
+      const user = userFabricator();
+      const createdUser = await createUser(user);
+
+      const searchResultPayload = searchResultFabricator(createdUser.id, 'node');
+      await createSearchResult(searchResultPayload);
+
+      await expect(findById(faker.string.uuid(), createdUser.id)).rejects.toHaveProperty(
```
tednguyendev, line 120: I think we don't need to use `faker` here, as we don't need this to be "realistic" , right?

tednguyendev (commented):
Just minor things 🐜

src/repositories/searchResult.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/78/commits/fbcf9f0cfb59da29becc014190b78bb67ef4aa64

test/services/searchResultEntity.test.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/78/commits/fbcf9f0cfb59da29becc014190b78bb67ef4aa64

test/services/searchResultEntity.test.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/78/commits/fbcf9f0cfb59da29becc014190b78bb67ef4aa64

Prabeshpd (commented):
I have added the fetch By Id to service file. Since, it has some custom errors to be raised. Is it okay to place it here? cc @olivierobert @tednguyendev

src/services/searchResultEntity.ts
```diff
@@ -0,0 +1,27 @@
...
+  if (!searchResult) {
+    const error = {
+      code: httpStatusCode.NOT_FOUND,
+      message: 'Keyword detail can not be found.',
+    };
+
+    throw error;
+  }
+
+  if (searchResult.userId !== userId) {
```
tednguyendev, line 17: Just minor: You can also use [Prisma's findMany](https://www.prisma.io/docs/concepts/components/prisma-client/crud#get-all-records), which will also allow you to pass the user.id to the query and we will be able to check both things(the newsletter id exists, and the user of the newsletter is the current user) 👀

An example: [Link](https://github.com/nimblehq/js-framework-benchmark/pull/126#discussion_r1228936098)

src/queries/searchResultQuery.ts
```diff
@@ -0,0 +1,31 @@
+class SearchResultQuery {
```
tednguyendev, line 1: This seems to conflict with the [the `getSearchResultQuery`](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/75/files#diff-90974662337520441e2025fcc59f378da40274248e5deb52452910029b78c41dR11)

test/services/searchResultEntity.test.ts
```diff
@@ -0,0 +1,77 @@
...
+describe('findById:', () => {
+  describe('given valid id', () => {
+    describe('given user has search result', () => {
+      it('returns valid data', async () => {
+        const createdUser = await createUser(user);
+
+        const searchResultPayload = [
+          {
+            ...searchResult,
+            id: 'cb77e671-2447-4197-897d-540bc114df48',
```
tednguyendev, line 22: I see we are using this `id` at many places. Maybe we should assign it to a variable ✋

tednguyendev (commented):
Another batch of minor things 🐜

src/services/searchResultEntity.ts
Prabeshpd: But doesn't it hides the fact that whether the user is `unauthorized` to view the result or the result does not exist entirely on the record (like deleted).

tednguyendev: Oh yes. We can not differentiate the 2 errors if we do it that way.

But since the logic of our application is that a user can only see his keywords, if one user is trying to view another user's keyword, I don't think it matters if we show what error to him, since he is trying to do the wrong, to access another person's data. We just need to prevent him.

Plus, if he wants to access the keyword of id `123`, and that's not his keyword, I still think it is also acceptable for the application to show the error "This keyword does not exist (to him)".

What do you think about this? 🤔

test/services/searchResultEntity.test.ts
```diff
@@ -0,0 +1,78 @@
...
+import { findById } from '@/services/searchResultEntity';
+import { createUser } from '@/repositories/user';
+import { createMultipleSearchResults } from '@/repositories/searchResult';
+
+import { user } from '../factories/user';
+import { searchResult } from '../factories/searchResult';
+
+describe('findById:', () => {
+  const defaultUUID = 'cb77e671-2447-4197-897d-540bc114df48';
+  describe('given valid id', () => {
```
tednguyendev, line 15: How about give the code a little space to breath 🌊

```suggestion
  const defaultUUID = 'cb77e671-2447-4197-897d-540bc114df48';
  
  describe('given valid id', () => {
```

Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/78/commits/e56350f43b948ee59561e6d3b21aed7291338acd

src/services/searchResultEntity.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/78/commits/e56350f43b948ee59561e6d3b21aed7291338acd

src/services/searchResultEntity.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import * as searchResultRepository from '@/repositories/searchResult';
+
+export async function findById(id: string, userId: number) {
+  const [searchResult] = await searchResultRepository.findById(id, userId);
+
+  if (!searchResult) {
+    const error = {
+      code: httpStatusCode.UNAUTHORIZED,
+      message: 'Unauthorized to view the keyword detail.',
```
tednguyendev, line 11: I think [what I meant](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/78#discussion_r1303886074) is that I'm more of an advocate for the 'Keyword detail can not be found.' if we can only choose 1 message to show 🤣 Do you find this 'Unauthorized' message more suitable? 🤔 

Anyway, I will approve this PR first so it does not block you ✋

tednguyendev (approved):

src/services/searchResultEntity.ts
Prabeshpd: Got your point. Will update the message. 😅

src/app/actions/searchResult.ts
```diff
@@ -30,3 +31,22 @@ export async function listSearchResults(paginationParams: PaginationParams, keyw
...
 }
+
+export async function findSearchResultById(id: string) {
+  return withAuth(async (currentUser) => {
+    try {
+      return findById(id, currentUser.id);
+    } catch (err: any) {
+      if (err.code) {
+        throw err;
+      }
```
olivierobert, line 42: This implementation is inconsistent with the method above. There is no extra logic to throw an exception in `listSearchResults`. Couldn't this be handled with:

```
code: error.code || httpStatusCode.UNPROCESSABLE_ENTITY,
```

src/app/actions/searchResult.ts
```diff
@@ -30,3 +31,22 @@ export async function listSearchResults(paginationParams: PaginationParams, keyw
...
+
+      const error = {
+        code: httpStatusCode.UNPROCESSABLE_ENTITY,
```
olivierobert, line 45: If a record cannot be fetched, then the error should always be 404 regardless of the error. Same as for `listSearchResults` too (this was missed before).

src/repositories/searchResult.ts
```diff
@@ -19,13 +19,16 @@ export type SearchResultScraping = Prisma.SearchResultGetPayload<{
   select: typeof searchResultScraping;
 }>;
 
-export const createSearchResult = async (
+export const createMultipleSearchResults = async (
```
olivierobert, line 22: How about being consistent with Prisma naming:

```suggestion
export const createManySearchResults = async (
```

olivierobert (changes requested):

src/repositories/searchResult.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/78/commits/bd13796cf9665b67851e198390d976e544820185

src/app/actions/searchResult.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/78/commits/ae63f9e8929978366a265fa00237061d9e3391c5

src/app/actions/searchResult.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/78/commits/ae63f9e8929978366a265fa00237061d9e3391c5

olivierobert (approved):

PR #84, opened 2023-08-28
[#18] [Backend] As a user, I can sign up with google

### LINKED ISSUE #18: [Backend] As a user, I can sign up with google
## Why
User may find it hassle to sign up with providing manual user name and password. So, we would like to use the OAUTH 2.0 strategy to leverage gmail to login in to the application.

## Acceptance Criteria

- User can use Google to sign up in the applications

Prabeshpd:
- Close #18

## What happened 👀

- Add identity table for storing google auth
- Add sign in option in next-auth
- Add a auth function for google auth

## Insight 📝
Used a separate identity table for storing the identity provider so the user can multiple sign in using different auth providers. Added the option in next-auth for credential as google OAuth. For users who have account registered but don't have provider setup would create a identity row on table for user. If the user don't have account it would create a account with random password and add identity.

## Proof Of Work 📹

Passed CI pipelines


prisma/migrations/20230828090447_identities/migration.sql
```diff
@@ -0,0 +1,14 @@
+-- CreateTable
+CREATE TABLE "identities" (
+    "id" UUID NOT NULL,
+    "provider" VARCHAR(100) NOT NULL,
+    "is_active" BOOLEAN NOT NULL DEFAULT false,
```
olivierobert, line 5: That flag does not seem used, hence could be removed

src/config/auth.ts
```diff
@@ -22,10 +23,25 @@ export const authOptions: AuthOptions = {
...
+      clientId: process.env.OAUTH_GOOGLE_CLIENT_ID as string,
+      clientSecret: process.env.OAUTH_GOOGLE_CLIENT_SECRET as string,
+    }),
   ],
   secret: process.env.NEXT_AUTH_SECRET,
   session: { strategy: 'jwt' },
   callbacks: {
+    async signIn({ account, profile }) {
+      if (account?.provider !== 'google') {
+        return true;
```
olivierobert, line 36: When does this case happens. I am confused 🤔

src/config/auth.ts
```diff
@@ -22,10 +23,25 @@ export const authOptions: AuthOptions = {
...
+      }
+
+      if (!profile) return false;
+
+      const user = await authenticateGoogleLogin(profile);
+
+      return !!user;
```
olivierobert, line 43: Avoud double negatives (https://dev.to/mcsee/code-smell-51-double-negatives-4fh0) as it is harder to comprehend and makes sense of It.

test/repositories/identity.test.ts
```diff
@@ -0,0 +1,75 @@
...
+ * @jest-environment node
+ */
+
+import { user } from '../factories/user';
+
+import * as userRepository from '@/repositories/user';
+import * as identityRepository from '@/repositories/identity';
+import { AuthProviders } from '@/constants/enum';
+
+describe('identity:', () => {
```
olivierobert, line 11: Naming issue:

```suggestion
describe('Identity Repository', () => {
```

test/repositories/identity.test.ts
```diff
@@ -0,0 +1,75 @@
...
+  describe('creates identity', () => {
```
olivierobert, line 12: Ditto as it should be the method name:

```suggestion
  describe('createIdentity', () => {
```

test/repositories/identity.test.ts
```diff
@@ -0,0 +1,75 @@
...
+          };
+          await identityRepository.createIdentity(identityPayload);
+
+          const [identity] = await identityRepository.findByProvider(AuthProviders.GOOGLE, userId);
+
+          expect(identity).not.toBeNull();
+        });
+      });
+
+      describe('given identity does not relates to user', () => {
```
olivierobert, line 49: English syntax issue + tip to emphasize the sad path:

```suggestion
      describe('given identity does NOT relate to user', () => {
```

test/repositories/identity.test.ts
```diff
@@ -0,0 +1,75 @@
...
+          await identityRepository.createIdentity(identityPayload);
+
+          const identity = await identityRepository.findByProvider(AuthProviders.GOOGLE, 2);
+
+          expect(identity.length).toEqual(0);
+        });
+      });
+    });
+
+    describe('given identity is not present', () => {
```
olivierobert, line 67: Minor reprhasing suggestion:

```suggestion
    describe('given identity does NOT exist', () => {
```

olivierobert (changes requested):

src/config/auth.ts
Prabeshpd: Sign in callback can activate for credentials too. So, I am returning it for the one where the provider is not google.

test/repositories/identity.test.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/84/commits/6b76e5248d44da9120517afdfbf615d3d1043d42

src/config/auth.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/84/commits/b90149544a7d35d98bfcb61fc30b6e34b93625fb

src/services/auth.ts
```diff
@@ -37,3 +45,41 @@ export async function authenticateLogin(credentials: Credentials) {
 
   return user;
 }
+
+export async function authenticateGoogleLogin(profile: Profile) {
```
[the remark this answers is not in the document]
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/84/commits/b90149544a7d35d98bfcb61fc30b6e34b93625fb

src/services/auth.ts
```diff
@@ -37,3 +45,41 @@ export async function authenticateLogin(credentials: Credentials) {
...
+  const [userIdentity] = await findByProvider(AuthProviders.GOOGLE, user?.id);
+  if (!userIdentity) {
+    const identityPayload = {
+      user: { connect: { id: user.id } },
+      provider: AuthProviders.GOOGLE,
+    };
+
+    await createIdentity(identityPayload);
+
+    return user;
```
tednguyendev: Will it still work if we remove this line? ✋

test/services/user.test.ts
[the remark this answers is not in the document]
Prabeshpd: As discussed in previous project I was trying to disintegrate the test of database using the real database instance. So, for those test I would require not to use mock database created.For this I had been using two config for jest to use what instance of database. A separate ticket has been created to address this issue. I am planning to have only the test which interact with database.

src/services/auth.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/84/commits/a9503b46947a434b871ec4651acd06e2462210fd

src/services/user.ts
```diff
@@ -0,0 +1,25 @@
...
+import httpStatusCode from 'http-status-codes';
+
+import * as crypt from '@/lib/crypt';
+import * as repositories from '@/repositories/user';
+
+export async function createUser(payload: Prisma.UserCreateInput) {
+  const user = await repositories.findByEmail(payload.email);
+  if (user) {
+    const error = {
+      code: httpStatusCode.BAD_REQUEST,
```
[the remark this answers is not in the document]
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/84/commits/a9503b46947a434b871ec4651acd06e2462210fd

prisma/schema.prisma
```diff
@@ -39,3 +40,15 @@ model SearchResult {
 
   @@map("search_results")
 }
+
+model Identity {
+  id           String                   @id @default(uuid()) @db.Uuid
+  provider     String                   @db.VarChar(100)
```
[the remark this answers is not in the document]
tednguyendev: Yes I don't mean to duplicate the enum.

What I mean is:

```suggestion
  provider     AuthProviders            @db.VarChar(100)
```

Can we use it like that 🤔 

Seems like Prisma Enum work with [key-value enum](https://stackoverflow.com/a/73569487) too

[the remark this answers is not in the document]
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/84/commits/8bf512a327a61e1d21c423084e0d6e1239d9a484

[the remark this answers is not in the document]
tednguyendev: But wait, with your implementation above, it will duplicate the enum? So Prisma not support us DRY this?

src/config/auth.ts
```diff
@@ -24,10 +26,25 @@ export const authOptions: AuthOptions = {
...
+    async signIn({ account, profile }) {
+      if (account?.provider === AuthProviders.CREDENTIALS) {
+        return true;
+      }
+
+      if (!profile) return false;
+
+      const user = await authenticateGoogleLogin(profile);
+
+      return Boolean(user);
```
tednguyendev, line 46: I still don't understand why the function `authenticateGoogleLogin` needs to return the `user` instead of just a boolean to indicate whether the authentication is successful or not.

src/services/user/user.ts
```diff
@@ -0,0 +1,63 @@
...
+    user: { email, name },
+    authProvider,
+  } = payload;
+
+  const password = generateRandomString(8);
+  const encryptedPassword = await crypt.hash(password);
+
+  const userPayload = { email, name, password: encryptedPassword };
+
+  return dbClient.$transaction(async (trx) => {
```
tednguyendev, line 42: What `trx` stands for? We should prefer using meaningful names instead of short variable names.

src/repositories/identity.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const createIdentity = async (
+  identityAttributes: Prisma.IdentityCreateInput,
+  trx?: Prisma.TransactionClient
+) => queryBuilder(trx).identity.create({ data: identityAttributes });
+
+export const findByProvider = async (
+  provider: AuthProviders,
+  userId: number,
+  trx?: Prisma.TransactionClient
```
tednguyendev, line 13: Why should we add this `trx` to the functions that not need to be executed in a transaction?

src/repositories/helper.ts
```diff
@@ -0,0 +1,5 @@
+import { Prisma } from '@prisma/client';
+
+import dbClient from '@/config/database';
+
+export const queryBuilder = (trx?: Prisma.TransactionClient) => trx || dbClient;
```
tednguyendev, line 5: I don't think this function actually "builds" anything, right? 😆

test/repository/identity.test.ts
```diff
@@ -0,0 +1,75 @@
...
+        };
+        const createdIdentity = await identityRepository.createIdentity(identityPayload);
+
+        expect(createdIdentity.userId).toEqual(userId);
+        expect(createdIdentity.provider).toEqual(AuthProviders.GOOGLE);
+      });
+    });
+  });
+
+  describe('find identity', () => {
```
tednguyendev, line 30:
```suggestion
  describe('findByProvider', () => {
```

As you are using this format above

test/services/auth.test.ts
```diff
@@ -2,59 +2,129 @@
...
+        expect(data.email).toEqual(user.email);
+        expect(data.name).toEqual(user.name);
       });
+    });
 
-      expect(data).not.toBeNull();
-      expect(data.email).toEqual(user.email);
-      expect(data.name).toEqual(user.name);
+    describe('when there is no user as credential', () => {
+      it('throws custom error.', async () => {
```
tednguyendev, line 33:
```suggestion
      it('throws custom error', async () => {
```

test/services/auth.test.ts
```diff
@@ -2,59 +2,129 @@
...
+        expect(data).not.toBeNull();
+        expect(data.email).toEqual(user.email);
+        expect(data.name).toEqual(user.name);
       });
+    });
 
-      expect(data).not.toBeNull();
-      expect(data.email).toEqual(user.email);
-      expect(data.name).toEqual(user.name);
+    describe('when there is no user as credential', () => {
```
tednguyendev, line 32:
```suggestion
    describe('given there is no user as credential', () => {
```

As you are using `given` many places

src/services/user/user.ts
Prabeshpd: will change it to transaction

src/config/auth.ts
Prabeshpd: So, must of our authenticate function normally returns user. I think it doesnot makes sense to return boolean for user authentication for the next-auth scenario. It's in service layer and can be used on other scenarios as well. For consistency I am returning the user.

src/repositories/helper.ts
Prabeshpd: renamed function name

prisma/schema.prisma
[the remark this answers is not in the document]
Prabeshpd: It seems the prisma does allow to use it in code. I have updated the PR.

src/config/auth.ts
tednguyendev: I see.

But what are the scenarios that this service will also be used? 🤔

src/repositories/identity.ts
tednguyendev: Thanks for your explanation.

Yes, if there are any functions that need to be used in a transaction, we can just add this additional param to them later 😄

src/config/auth.ts
Prabeshpd: Since, We are building API for OAUTH. Although there is a API from next-auth but I think it is little complicate to extract token by clients we may want to expose the API for mobile devices separately. So, I was thinking to make separate API. But if we do not use this approach too. I think it is not good to return boolean from the authenticate function. I may be thinking too much. 😅 . I think we can create a separate function `hasAuthenticatedGoogleLogin` for boolean. Do you think is it a good idea? to define separate function

tednguyendev: Ohh I see.

Yes, then maybe we can just leave it as `authenticateGoogleLogin` for now. We can comeback to this later.

tednguyendev (approved):

src/services/user/user.ts
```diff
@@ -0,0 +1,63 @@
...
+      provider: authProvider,
+    };
+
+    await createIdentity(identityPayload, transaction);
+
+    return createdUser;
+  });
+}
+
+export async function mergeExitingUser(user: User, authProvider: AuthProviders) {
```
olivierobert, line 55: Typo:

```suggestion
export async function mergeExistingUser(user: User, authProvider: AuthProviders) {
```

src/config/auth.ts
olivierobert: It would be best to then have a method lie the following:

```
if (isAuthCredentialFlow(account?.provider)) {
        return true;
}
```

Such code is self-documenting and it helps to understand what is going on 💡

src/services/auth/auth.ts
```diff
@@ -0,0 +1,57 @@
...
+
+import AuthError from './error';
+import { createUserWithIdentity, mergeExitingUser } from '../user/user';
+
+interface Profile {
+  email?: string;
+  name?: string;
+}
+
+export async function authenticateLogin(credentials: Credentials) {
```
olivierobert, line 16: The method name is hard to differentiate (especially when checking the tests). How about using the word `credential` as `Google` is used in the second method:

```suggestion
export async function authenticateCredentialLogin(credentials: Credentials) {
```

test/services/auth.test.ts
```diff
@@ -2,59 +2,128 @@
...
 describe('Auth:', () => {
-  describe('given correct credentials', () => {
-    it('returns user', async () => {
-      const userPayload = { name: user.name, email: user.email, password: user.password };
-
-      await createUser(userPayload);
-      const data = await authenticateLogin({
-        email: user.email,
-        password: user.password,
+  describe('Authenticate credential login', () => {
```
olivierobert, line 14: Inconsistency with the other tests. Prefer to use the method name:

```suggestion
  describe('authenticateLogin', () => {
```

test/services/auth.test.ts
```diff
@@ -2,59 +2,128 @@
...
-        authenticateLogin({
-          email: 'user@nimblehq.co',
-          password: '1234567',
-        })
-      ).rejects.toHaveProperty('code', 400);
+  describe('Authenticate google login', () => {
+    describe('given valid profile with no previous user', () => {
+      it('returns user after creating user and identity', async () => {
+        const userEmail = 'dev@nimblehq.co';
+        const profile = { email: userEmail, name: 'dev nimble' };
```
olivierobert, line 71: As raised previously a test should follow the AAA pattern. So let's add a line break here after the Arrange block:

```suggestion
        const profile = { email: userEmail, name: 'dev nimble' };
        
```

Then the act block should be (one line):

```js
await authenticateGoogleLogin(profile);
```

Finally, the Assert block is:

```js
const authenticateUser = (await userRepository.findByEmail(userEmail)) as User;
const [identity] = await identityRepository.findByProvider(
  AuthProviders.google,
  authenticateUser.id
);

expect(identity).not.toBeNull();
expect(authenticateUser.email).toEqual(userEmail);
```

You need to recheck all tests added in this PR.

olivierobert (changes requested):

test/services/auth.test.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/84/commits/f3a388edb96ff807b910abc71110e3dcca753c4b

test/services/auth.test.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/84/commits/f3a388edb96ff807b910abc71110e3dcca753c4b

olivierobert (approved):

PR #85, opened 2023-08-29
Feature/UI google sign in

### LINKED ISSUE #17: [UI] As a user, I can sign up with google
## Why

User may want to login using the gmail. So. for this a login button has to be created for signing up with google.

## Acceptance Criteria

- Create a button for signing up with google
- Once clicked send the request to backend

Prabeshpd:
- Close #17 

## What happened 👀

Provide a description of the **changes** this pull request brings to the codebase. Additionally, when the pull request is still being worked on, a checklist of the planned changes is welcome to track progress.

## Insight 📝

Describe in detail why this solution is the most appropriate, which solution you tried but did not go with, and how to test the changes. References to relevant documentation are welcome as well.

## Proof Of Work 📹

Show us the implementation: screenshots, GIFs, etc.


tednguyendev:
Woa very concise PR 😆 Are you only using 1 Google account, so you don't need to choose accounts after clicking the sign in button?

src/app/auth/login/page.tsx
```diff
@@ -29,10 +29,22 @@ const Login = () => {
...
+      toast(err.message, 'error');
+    }
+  };
+
   return (
     <main className="layout-auth">
       <h2 className="layout-auth__heading">Log In Form</h2>
       <LogInForm handleFormSubmit={handleFormSubmit} />
+      <button onClick={signInGoogle} className="form__button--submit">
+        Sign In with Google
```
tednguyendev, line 46: As you are using "Login", maybe we can make it 

```suggestion
        Log In with Google
```

CleanShot 2023-08-30 at 17 32 59@2x

tednguyendev (approved):

src/app/auth/login/page.tsx
```diff
@@ -29,10 +29,22 @@ const Login = () => {
...
+    } catch (err: any) {
+      toast(err.message, 'error');
+    }
+  };
+
   return (
     <main className="layout-auth">
       <h2 className="layout-auth__heading">Log In Form</h2>
       <LogInForm handleFormSubmit={handleFormSubmit} />
+      <button onClick={signInGoogle} className="form__button--submit">
```
olivierobert, line 45: Thats an incorrect use of BEM. Classes must be as the following:

```suggestion
      <button onClick={signInGoogle} className="form__button form__button--submit">
```

The first class allows to inherit the general styles for `form__button` while the second class applies the variant styles for `form__button--submit`.

olivierobert (changes requested):

src/app/auth/login/page.tsx
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/85/commits/4cf03096935de24b8d3cb3809ceb398658818661

olivierobert (approved):

PR #93, opened 2023-08-30
[#90] [API] As a user, I can view the list of search results

### LINKED ISSUE #90: [API] As a user, I can view the list of search results
## Why

 A client who may be mobile user wants to use the services that have been developed. API are needs to be made for viewing search result so that they can be consumed by client

## Acceptance Criteria

- Create an API to create user.
- API must have query to support pagination and search feature
- Add test

Prabeshpd:
- Close #90 

## What happened 👀

- Created a api to view search results list
- Add tests
- Created a presentation function to wrap the data and meta in api response

## Insight 📝
Created an api to view the search result list. For it query params is used to take the pagination and filter params. After the query a presentation function is created. Defining the structure of an API in controller or route pollutes them. I wanted to have a separate place for them and borrowed the idea from ruby.

## Proof Of Work 📹

Passed CI pipelines

https://github.com/nimblehq/ic-nextjs-prabesh-dahal/assets/39412015/fa7decd8-3a06-4692-815c-0ba3b8f357d6


src/app/api/v1/searchResults/route.ts
```diff
@@ -0,0 +1,45 @@
...
+    const userId = Number(url.searchParams.get('userId') || 1);
+
+    const paginationQuery = getPaginationParams({ currentPage, maxRows });
+    const customQuery = getSearchResultQuery({ userId, keyword });
+
+    const searchResults = await listSearchResult({
+      customQuery,
+      paginationQuery,
+    });
+    const totalRowCount = (await listSearchResult({ customQuery }, { isPaginated: false })).length;
```
tednguyendev, line 27: This looks duplicated with the [searchResult.ts action](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/blob/b45084acb3244d9e6fe9894f6195109cd871aa82/src/app/actions/searchResult.ts#L14-L21).

test/api/searchResult.test.ts
```diff
@@ -0,0 +1,148 @@
...
+import { NextRequest, NextResponse } from 'next/server';
+
+import { GET } from '@/app/api/v1/searchResults/route';
+import { createManySearchResults } from '@/repositories/searchResult';
+import { createUser } from '@/repositories/user';
+
+import { user } from '../factories/user';
+import { searchResult } from '../factories/searchResult';
+
+describe('search result Get API:', () => {
```
tednguyendev, line 11: Should this be

```suggestion
describe('Search result get API:', () => {
```

As you are using [that format](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/blob/3fdabf9f5438a15d6eb9e466d0f1fafde1f18e95/test/api/user.test.ts#L9) 🧓

src/presenters/searchResult.ts
```diff
@@ -0,0 +1,7 @@
+import { SearchResult } from '@prisma/client';
+
+import { PaginatedMeta } from '@/lib/pagination';
+
+export const searchResultListPresenter = (data: SearchResult[], meta: PaginatedMeta) => {
+  return { searchResults: data, meta };
```
tednguyendev, line 6: If our presenter does not do any transformations, I wonder if we need it.

This is not something that pollutes the controller 😆

src/app/api/v1/searchResults/route.ts
Prabeshpd: Yes, As the one in action is for the server actions. It will have its own authentication function. So, these two lines are repeated in API

src/presenters/searchResult.ts
Prabeshpd: For, now I am not doing anything but to present in table I am thinking to have the map to format the data a little. I will have a ticket on next sprint to add columns in table.

src/app/api/v1/searchResults/route.ts
tednguyendev: As the query will change, maybe we can skip it for now 🙏

src/presenters/searchResult.ts
tednguyendev: Since you already added it, just leave it there 👍 

But I think we should not have the "preparation" code like this too much, since although they are harmless, they may cause some confusion for the reviewers(like me 🙈 )

test/api/searchResult.test.ts
```diff
@@ -0,0 +1,148 @@
...
+
+describe('Search result get API:', () => {
+  describe('given only pagination params', () => {
+    it('returns all the data with pagination', async () => {
+      const createdUser = await createUser(user);
+
+      const searchResultPayload = [
+        {
+          ...searchResult,
+          id: 'cb77e671-2447-4197-897d-540bc114df48',
```
olivierobert, line 19: Since there is no assertion on the ID, can't this be removed and instead rely on the factory? It adds noise to the test for no benefit.

test/api/searchResult.test.ts
```diff
@@ -0,0 +1,148 @@
+import { createMocks } from 'node-mocks-http';
+import { NextRequest, NextResponse } from 'next/server';
+
+import { GET } from '@/app/api/v1/searchResults/route';
+import { createManySearchResults } from '@/repositories/searchResult';
+import { createUser } from '@/repositories/user';
+
+import { user } from '../factories/user';
+import { searchResult } from '../factories/searchResult';
```
olivierobert, line 9: Let's be more descriptive as I wondered where `searchResult` came from when reading the code below:

```suggestion
import { searchResultFactory } from '../factories/searchResult';
```

test/api/searchResult.test.ts
```diff
@@ -0,0 +1,148 @@
...
+      expect(data.meta).toEqual({
+        totalRowCount: 2,
+        totalPageCount: 1,
+        perPageCount: 10,
+        currentPage: 1,
+      });
+    });
+  });
+
+  describe('given pagination and filter params', () => {
```
olivierobert, line 58: Tehnically, this test does not add value as it is the same as the other test. To make this test more useful, we would need to test to catch with different pagination pages (e.g., currentPage = 2)

src/app/api/v1/searchResults/route.ts
olivierobert, line 1: URLs must be hyphenated, no camel-cased:

```
src/app/api/v1/search-results/route.ts
```

test/api/searchResult.test.ts
```diff
@@ -0,0 +1,148 @@
...
+import { NextRequest, NextResponse } from 'next/server';
+
+import { GET } from '@/app/api/v1/searchResults/route';
+import { createManySearchResults } from '@/repositories/searchResult';
+import { createUser } from '@/repositories/user';
+
+import { user } from '../factories/user';
+import { searchResult } from '../factories/searchResult';
+
+describe('Search result get API:', () => {
```
olivierobert, line 11: Keep it simple with:

```suggestion
describe('GET api/v1/search-results', () => {
```

olivierobert (changes requested):

test/api/searchResult.test.ts
Prabeshpd: Since, if I change this naming I have to change in all of the files that use it. Would it be good if I do this in the task where I move all the tests to one directory.

test/api/searchResult.test.ts
Prabeshpd: Since, the searchResult is an object in the factory file. It creates a single ID as UUID. If we try to insert multiple searchResults with same UUID we can not as it is a unique field.

test/api/searchResult.test.ts
```diff
@@ -0,0 +1,148 @@
...
+      await createManySearchResults(searchResultPayload);
+
+      let { req }: { req: NextRequest; res: NextResponse } = createMocks({
+        method: 'GET',
+        headers: { 'content-type': 'application/json' },
+      });
+
+      req = {
+        ...req,
+        url: `http://localhost:3000/api/v1/searchResults?currentPage=1&maxRows=10&userId=${createdUser.id}`,
```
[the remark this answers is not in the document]
Prabeshpd: No, It is not possible. As per my knowledge and research I did Next Request does not have the native query object. The code have to extract the query params from the url. Due to this we can not add the query object and add the params in the api url itself.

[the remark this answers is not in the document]
olivierobert: I did not test it but I found many examples showing support for `query`:

Example 1: https://seanconnolly.dev/unit-testing-nextjs-api-routes

image

Example 2: https://www.paigeniedringhaus.com/blog/how-to-unit-test-next-js-api-routes-with-typescript

image

In all examples, I noticed that it was to set the value for the path of the URL e.g., `'/api/[animal]'`. So which issue did you face?

Also, is it necessary to set the host and port? Can't `localhost:3000` be removed?

test/api/searchResult.test.ts
olivierobert: Understood. Then ensure you keep track of these changes to make.

test/api/searchResult.test.ts
olivierobert: But isn't the point of using Faker to generate random UUID? 🤔 Having to define it by hand in each test seems to defeat the point of having a factory.

https://github.com/nimblehq/ic-nextjs-prabesh-dahal/blob/73ae713239079f3716a5427df9f59da853aa99fb/test/factories/searchResult.ts#L5

test/api/searchResult.test.ts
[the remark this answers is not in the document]
Prabeshpd: I tried it but the next Request is not taking the query. I think it's for next-12. I will give it a try once more. May be I had missed anything.

test/api/searchResult.test.ts
Prabeshpd: Since, it is a object. The import instance has the object instance defined once. One thing we can do is call ```faker.string.uuid()``` in every object assign operator. I will do that.

src/app/api/v1/search-results/route.ts
```diff
@@ -0,0 +1,46 @@
...
+export async function GET(request: Request) {
+  try {
+    const url = new URL(request.url);
+
+    const keyword = url.searchParams.get('keyword');
+    const currentPage = Number(url.searchParams.get('currentPage') || PAGINATION_CURRENT_PAGE);
+    const maxRows = Number(url.searchParams.get('maxRows') || PAGINATION_LIMIT);
+
+    // TODO: Remove after api auth
+    const userId = Number(url.searchParams.get('userId') || 1);
```
tednguyendev, line 19: We will also remove this in this PR?

Prabeshpd: I don't think. Since it is a separate implementation for api authentication. It's safe to remove them on the separate PR. cc @olivierobert

test/api/searchResult.test.ts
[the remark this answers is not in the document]
Prabeshpd: Here is the[Link](https://dev.to/bikrantjung/how-to-access-url-params-in-incoming-request-nextjs-134-route-handlers-2887). I believe the nextjs 13 does not have the req.query context. So, we need to take the params for url. We cannot remove the host and port as we are converting the url from the request to URL object which will not be valid.

test/api/searchResult.test.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/93/commits/2b5e36f1c38ef3094d28d911e9b96c32a94d9266

test/api/searchResult.test.ts
```diff
@@ -0,0 +1,143 @@
...
+        },
+      ];
+      await createManySearchResults(searchResultPayload);
+      let { req }: { req: NextRequest; res: NextResponse } = createMocks({
+        method: 'GET',
+        headers: { 'content-type': 'application/json' },
+      });
+      req = {
+        ...req,
+        url: `http://localhost:3000/api/v1/searchResults?currentPage=1&maxRows=10&userId=${createdUser.id}`,
```
tednguyendev, line 38: Ditto [this](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/94#discussion_r1315554299). And other places in this test too 👍

test/api/searchResult.test.ts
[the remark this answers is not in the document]
olivierobert: Then let's file it under a known issue of Next.JS 13. In that case, I recommend defining a constant or an env variable (used only in the test environment) to store `http://localhost:3000` so the test looks like:

```js
url: `http://${TEST_APP_HOST}/api/v1/searchResults?currentPage=1&maxRows=10&userId=${createdUser.id}`
```

It would allow not to duplicate this code, allows to change based on the environment but also add a comment where the contact or env is configured to document why this is needed:

```
// Request  in Next.JS > 13 does not support the query attribute, hence requires a fully qualified URL
TEST_APP_HOST="http://localhost:3000"
```

The implementation should be self-documenting the decisions made so that others will not have to figure out again the same challenges.

olivierobert (approved):

test/api/searchResult.test.ts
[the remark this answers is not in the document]
Prabeshpd: Have added the constant in env variable

tednguyendev (approved):

PR #101, opened 2023-09-04
[#20] [Backend] As a user, I can search for certain word across url and keyword

### LINKED ISSUE #20: [Backend] As a user, I can search for certain word across adword URLS
## Why

User may need to know whether certain word is there in URLS or not.

## Acceptance Criteria

- Create a function to search for certain word across adword URLS using regex.
- Add test

Prabeshpd:
- Close #20 

## What happened 👀

- Add filter logic to search results for certain word in url and keyword 
- Discern the query with search scope and type
- Add test

## Insight 📝

Added filter logic to get the search result for searching the url and keyword. Scope parameter is used to discern whether the search is for keyword or url. Type parameter is used to discern whether the query should be full search or partial search.

## Proof Of Work 📹

Passing CI pipelines


src/queries/SearchResultQuery.ts
```diff
@@ -13,26 +13,30 @@ interface SearchResultQueryInterface extends BaseQueryInterface<SearchResult> {
...
     this.url = queryParams.filters?.url;
+    this.filterAdUrlWord = queryParams.filters?.filterAdUrlWord;
   }
 
   async call() {
     this.buildUserQueryParams();
     this.buildKeywordQueryParams();
     this.buildUrlToFilterQueryParams();
     this.buildPaginationQueryParams();
+    await this.buildUrlWithWordFilterQueryParams();
```
tednguyendev, line 39: We can rearrange the build functions so that it build the `this.filterQuery` first, `this.paginationQuery` later

```suggestion
    this.buildUserQueryParams();
    this.buildKeywordQueryParams();
    this.buildUrlToFilterQueryParams();
    await this.buildUrlWithWordFilterQueryParams();
    this.buildPaginationQueryParams();
```

src/queries/SearchResultQuery.ts
```diff
@@ -63,6 +67,30 @@ class SearchResultQuery extends BaseQuery<SearchResult> implements SearchResultQ
...
+    const validIds = searchResults
+      .filter((searchResult: SearchResult) =>
+        searchResult.adWordTopUrls.join(',').includes(wordToSearch)
+      )
+      .map((searchResult: SearchResult) => searchResult.id);
+
+    const existingFilterIds = this.filterQuery.id?.in ?? [];
+    this.filterQuery['id'] = {
+      in: [...existingFilterIds, ...validIds],
+    };
```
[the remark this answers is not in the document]
Prabeshpd: There is another filter that also sets the id property in filter. So, I am making sure that the existing ids don't get override and combining both array of ids.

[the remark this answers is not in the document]
tednguyendev: Ah yes I just saw [that](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/102/files#diff-9649ebec8ad27f7ce281f313ccbda1b813badaffaa600c0a3388722e6e76060eR102) too. We can put this into the later PRs 🙈 

But as I see, the `buildUrlContainsAtLeastSpecificCharactersQueryParams` is executed after `buildUrlWithWordFilterQueryParams`. So do we need to worry about the override here? Should we only check the override in the `buildUrlContainsAtLeastSpecificCharactersQueryParams`?

test/queries/searchResultQuery.test.ts
```diff
@@ -133,7 +133,85 @@ describe('SearchResultQuery', () => {
...
+
+        const data = await searchResultQuery.call();
+
+        expect(data.length).toEqual(3);
+      });
+    });
+  });
+
+  describe('given ad words url param', () => {
+    describe('given valid param for filter ad words url containing word', () => {
```
tednguyendev, line 169:
```suggestion
    describe('given there is filterAdUrlWord filter, () => {
```

test/queries/searchResultQuery.test.ts
```diff
@@ -133,7 +133,85 @@ describe('SearchResultQuery', () => {
...
+          filters: { filterAdUrlWord: 'nimble' },
+        });
+
+        const data = await searchResultQuery.call();
+
+        expect(data.length).toEqual(1);
+      });
+    });
+
+    describe('there is NO filter ad words url param', () => {
```
tednguyendev, line 208:
```suggestion
    describe('given there is NO filterAdUrlWord filter', () => {
```

src/queries/SearchResultQuery.ts
```diff
@@ -60,6 +61,29 @@ class SearchResultQuery extends BaseQuery<SearchResult> implements SearchResultQ
       },
     ];
   }
+
+  private async buildUrlWithWordQueryParams() {
+    if (!this.filters?.filterAdUrlWord) return;
```
tednguyendev, line 66: The name `filterAdUrlWord` sounds duplicated(the `filter` repeat twice, `filters?.filterAdUrlWord`), along with [other filters](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/102/files#diff-a19c21310c5cc89a48f322e52cc9e56d4307d265323e76416f6f1c2ee43ef2cbR5-R6).

How about just giving this the name `wordToSearch` like how you are doing in [the line under this](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/101/files#diff-9649ebec8ad27f7ce281f313ccbda1b813badaffaa600c0a3388722e6e76060eR68)?

src/types/query.ts
```diff
@@ -1,4 +1,5 @@
 export interface SearchResultFilter {
   keyword?: string | null;
   url?: string | null;
+  wordToSearch?: string | null;
```
olivierobert, line 4: Since the search is either on the keyword or URL there could be only two params. In terms of UX, a users can either do:

- keyword = "test" (exact or partial match)
- URL = "google" (partial match), "www.google.com" (exact match)

src/queries/SearchResultQuery.ts
```diff
@@ -60,6 +61,29 @@ class SearchResultQuery extends BaseQuery<SearchResult> implements SearchResultQ
       },
     ];
   }
+
+  private async buildUrlWithWordQueryParams() {
+    if (!this.filters?.wordToSearch) return;
+
+    const wordToSearch = this.filters.wordToSearch;
+    const searchResults = await this.databaseModel.findMany({ where: { userId: this.userId } });
```
olivierobert, line 69: Having to do a query before the main query is not ideal. Technically that should be done as a subquery. Did you consider using raw queries: https://www.prisma.io/docs/concepts/components/prisma-client/raw-database-access?

Prabeshpd: Yes, I did consider using raw queries. Upon discussing with developers everyone suggested to not to go with raw query. So, went with this approach.

src/queries/SearchResultQuery.ts
```diff
@@ -26,42 +33,76 @@ class SearchResultQuery extends BaseQuery<SearchResult> implements SearchResultQ
...
+      where: {
+        keyword: { contains: keyword ?? '' },
+        OR: [
+          {
+            adWordTopUrls: {
+              has: text,
+            },
+          },
+          {
+            nonAdUrls: {
```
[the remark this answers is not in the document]
Prabeshpd: That is covered in previous PR that is merged. Since, there was code structure change had to change it too. 🙂

src/components/SearchResultList/SearchResultList.tsx
```diff
@@ -45,10 +45,10 @@ const SearchResultList = () => {
     setSearchKeyWord(event.target.value);
   };
 
-  const listSearchResults = async (queryParams: PaginationParams, keyword: string) => {
+  const listSearchResults = async (paginationParams: PaginationParams, keyword: string) => {
     setLoading(true);
     try {
-      const data = await listSearchResultsAction(queryParams, { keyword });
+      const data = await listSearchResultsAction(paginationParams, { keyword });
```
olivierobert, line 51: We missed that variable naming but usually avoid using `data`:

```suggestion
      const fetchedSearchResult = await listSearchResultsAction(paginationParams, { keyword });
```

src/queries/SearchResultQuery.ts
```diff
@@ -26,42 +33,76 @@ class SearchResultQuery extends BaseQuery<SearchResult> implements SearchResultQ
...
+      ...this.paginationQuery,
+    });
+  }
+
+  private async filterByPartialUrl(text: string, keyword: string | null) {
+    return dbClient.$queryRaw(
+      Prisma.sql`SELECT * FROM search_results sr
+       WHERE sr."userId" = ${this.userId}
+       AND sr."keyword" ~ ${keyword ?? ''}
+       AND array_to_string(sr.ad_word_top_urls, ', ') ~ ${text}
```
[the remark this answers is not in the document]
Prabeshpd: Oh. It seems I have interchanged pattern and partial. Will update it. Thank you

src/queries/SearchResultQuery.ts
```diff
@@ -26,42 +33,76 @@ class SearchResultQuery extends BaseQuery<SearchResult> implements SearchResultQ
...
 
-  private buildUserQueryParams() {
-    this.filterQuery['userId'] = this.userId;
-  }
+  private async filterUrl(urlFilterParams: UrlFilterParams) {
+    const { text, keyword, searchType } = urlFilterParams;
 
-  private buildKeywordQueryParams() {
-    if (!this.filters?.keyword) return;
+    switch (searchType) {
```
olivierobert, line 68: I do not see a valid reason why the search type would apply to only to matching URLs. In terms of UX, I would expect that the search type would apply to both: I want to match the exact keyword or the exact URL at the same time.

src/queries/SearchResultQuery.ts
```diff
@@ -26,42 +33,76 @@ class SearchResultQuery extends BaseQuery<SearchResult> implements SearchResultQ
...
+
+    return this.filterUrl(urlFilterParams);
+  }
+
+  private async filterByKeyword(keyword: string | null) {
     this.buildPaginationQueryParams();
 
     return this.databaseModel.findMany({
-      where: { ...this.filterQuery },
+      where: { userId: this.userId, keyword: { contains: keyword ?? '' } },
```
olivierobert, line 60: The fallback to an empty string seems redundant if  `filterByKeyword` was only called if `keyword` was not null.

src/queries/SearchResultQuery.ts
```diff
@@ -26,42 +33,76 @@ class SearchResultQuery extends BaseQuery<SearchResult> implements SearchResultQ
...
+
+    if (!this.filters) {
+      return this.databaseModel.findMany({
+        where: { userId: this.userId },
+        ...this.paginationQuery,
+      });
+    }
+
+    const { keyword = '' } = this.filters;
+    if (!this.filters.url?.text) {
```
olivierobert, line 46: This is quite expected because on the previous line we check for the value of the keyword and then the conditional is for checking if the value of the URL is set 🤔

src/queries/SearchResultQuery.ts
Prabeshpd: Since, keyword and url are different terms. I was planning to implement two different search for it. User may wants to search only the keywords without searching for the URLs. I was thinking since those two are separate concerns it would make sense to treat them differently. However, I have applied AND filter with keyword and filter params.

src/queries/SearchResultQuery.ts
Prabeshpd: Empty string are being treated as null in typescript. That's why I am using a fallback. Prisma appraently don't treat null as empty string.

src/queries/SearchResultQuery.ts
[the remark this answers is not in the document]
tednguyendev: So if we do exact search, we also search nonAdUrls, but we don't need to search nonAdUrls in partial search? 👀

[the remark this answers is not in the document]
Prabeshpd: Yes, That is the requirement in IC. 😅

src/queries/SearchResultQuery.ts
```diff
@@ -26,42 +32,98 @@ class SearchResultQuery extends BaseQuery<SearchResult> implements SearchResultQ
...
+        ...this.paginationQuery,
+      });
+    }
+
+    const { text, searchElement, searchType } = this.filters;
+
+    return this.filterResults({ text, searchElement, searchType });
+  }
+
+  private async filterByKeyword(keyword: string): Promise<SearchResult[]> {
```
olivierobert, line 49: This method could be moved below `filterResults` as it is being used by that method so it males the file easier to read

src/queries/SearchResultQuery.ts
```diff
@@ -26,42 +32,98 @@ class SearchResultQuery extends BaseQuery<SearchResult> implements SearchResultQ
...
+      case SearchElement.KEYWORD:
+        return this.filterByKeyword(text);
 
-  private buildKeywordQueryParams() {
-    if (!this.filters?.keyword) return;
+      case SearchElement.URL:
+        return this.filterUrl({ text, searchType });
 
-    this.filterQuery['keyword'] = { contains: this.filters?.keyword };
+      case SearchElement.ALL:
```
olivierobert, line 72: This should the default case

src/queries/SearchResultQuery.ts
```diff
@@ -26,42 +32,98 @@ class SearchResultQuery extends BaseQuery<SearchResult> implements SearchResultQ
...
+    if (!this.filters?.text) {
+      return this.databaseModel.findMany({
+        where: { userId: this.userId },
+        ...this.paginationQuery,
+      });
+    }
+
+    const { text, searchElement, searchType } = this.filters;
+
+    return this.filterResults({ text, searchElement, searchType });
```
olivierobert, line 46: How about:

```suggestion
    return this.filterResults({ text, scope, type });
```

src/queries/SearchResultQuery.ts
```diff
@@ -26,42 +31,144 @@ class SearchResultQuery extends BaseQuery<SearchResult> implements SearchResultQ
...
 
-    this.filterQuery['keyword'] = { contains: this.filters?.keyword };
+      case SearchType.PATTERN:
+        return this.filterByPatternUrl(text);
+
+      default:
+        return this.databaseModel.findMany({
+          where: { userId: this.userId, keyword: { contains: text } },
+          ...this.paginationQuery,
+        });
```
tednguyendev, line 90: This should query the URL fields, not keyword, right? 🙈

src/queries/SearchResultQuery.ts
```diff
@@ -26,42 +31,144 @@ class SearchResultQuery extends BaseQuery<SearchResult> implements SearchResultQ
...
+
+  private async filterResults(filter: FilterParams): Promise<SearchResult[]> {
+    const { text = '', scope = SearchScope.ALL, type = SearchType.PARTIAL } = filter;
+
+    switch (scope) {
+      case SearchScope.KEYWORD:
+        return this.filterByKeyword(text, type);
+
+      case SearchScope.URL:
+        return this.filterUrl({ text, type });
```
tednguyendev, line 56:
```suggestion
        return this.filterByUrl({ text, type });
```

src/queries/SearchResultQuery.ts
```diff
@@ -26,42 +31,144 @@ class SearchResultQuery extends BaseQuery<SearchResult> implements SearchResultQ
...
+
+      default:
+        return this.filterUrlAndKeyword({ text, type });
```
tednguyendev, line 59:
```suggestion
        return this.filterByUrlAndKeyword({ text, type });
```

src/queries/SearchResultQuery.ts
```diff
@@ -26,42 +31,144 @@ class SearchResultQuery extends BaseQuery<SearchResult> implements SearchResultQ
...
-          has: this.filters?.url,
-        },
+      case SearchType.PATTERN:
+        return this.filterByPatternKeywordAndUrl(text);
+
+      default:
+        return this.databaseModel.findMany({
+          where: { userId: this.userId, keyword: { contains: text } },
+          ...this.paginationQuery,
+        });
```
tednguyendev, line 108: Same question: Shouldn't this also query URL fields?

src/queries/SearchResultQuery.ts
```diff
@@ -26,42 +31,144 @@ class SearchResultQuery extends BaseQuery<SearchResult> implements SearchResultQ
...
   }
 
-  private buildUserQueryParams() {
-    this.filterQuery['userId'] = this.userId;
+  private async filterByKeyword(keyword: string, type: SearchType): Promise<SearchResult[]> {
+    switch (type) {
+      case SearchType.EXACT:
+        return this.databaseModel.findMany({ where: { userId: this.userId, keyword } });
+
+      default:
```
tednguyendev, line 68: Is it the requirement that we don't have the pattern search type for keyword 😵

Prabeshpd: For keyword there is no requirement to have a pattern match. Requirement only focus on the url. For keyword it's either exact or similar words. contains query takes place of it.

src/queries/SearchResultQuery.ts
Prabeshpd: It's the fallback if anything comes that match the condition. It should not just keeping a default value if invalid type comes in to return basic filtered data.

src/queries/SearchResultQuery.ts
olivierobert: Then it should be a if/else block not a switch case. We could also extend to support pattern for any scope.

src/queries/SearchResultQuery.ts
olivierobert: That would be totally an unexpected side effect. The method is not doing what is says 😨

src/queries/SearchResultQuery.ts
```diff
@@ -26,42 +31,144 @@ class SearchResultQuery extends BaseQuery<SearchResult> implements SearchResultQ
...
-        },
+      ...this.paginationQuery,
+    });
+  }
+
+  private async filterByExactKeywordAndUrl(text: string): Promise<SearchResult[]> {
+    return this.databaseModel.findMany({
+      where: {
+        userId: this.userId,
+        keyword: text ?? '',
```
olivierobert, line 137: If this method is called, `text` should not /must not be an empty string

src/queries/SearchResultQuery.ts
Prabeshpd: I have moved the pattern search to the default one. And removed this.

src/queries/SearchResultQuery.ts
```diff
@@ -26,42 +32,194 @@ class SearchResultQuery extends BaseQuery<SearchResult> implements SearchResultQ
...
     this.buildPaginationQueryParams();
 
+    if (!this.filters?.queryInput) {
+      return this.databaseModel.findMany({
+        where: { userId: this.userId },
+        ...this.paginationQuery,
+      });
+    }
+
+    const { queryInput, type, scope, leastCount = 1 } = this.filters;
```
olivierobert, line 44: Setting the default value for `leastCount` is rather unexpected 🤔 Shouldnt this be defined at the interface level or in `filterResults` instead?

As a side note, `leastCount` does not seem used in the queries at all so could be dropped. to keep things simple (the queries are already complex enough).

src/queries/SearchResultQuery.ts
```diff
@@ -26,42 +32,194 @@ class SearchResultQuery extends BaseQuery<SearchResult> implements SearchResultQ
...
+    return dbClient.$queryRaw(
+      Prisma.sql`
+      SELECT flatten.id, flatten.keyword, flatten.nod_ad_urls, flatten.ad_word_top_urls, flatten.status  FROM 
+        (SELECT sr.id,
+          sr.keyword,
+          UNNEST(sr.nod_ad_urls) as ad_url_string, 
+          UNNEST(sr.ad_word_top_urls) as non_ads_url_string, 
+          sr.nod_ad_urls, 
+          sr.ad_word_top_urls,
+          sr.status
```
olivierobert, line 187: Why is status needed?

src/queries/SearchResultQuery.ts
```diff
@@ -26,42 +32,194 @@ class SearchResultQuery extends BaseQuery<SearchResult> implements SearchResultQ
...
+          sr.keyword,
+          UNNEST(sr.nod_ad_urls) as ad_url_string, 
+          UNNEST(sr.ad_word_top_urls) as non_ads_url_string, 
+          sr.nod_ad_urls, 
+          sr.ad_word_top_urls,
+          sr.status  
+          FROM search_results sr
+          WHERE sr."userId"=${this.userId}
+        ) AS flatten
+      where
```
olivierobert, line 191: Casing inconsistency:

```suggestion
      WHERE
```

src/queries/SearchResultQuery.ts
```diff
@@ -26,42 +32,194 @@ class SearchResultQuery extends BaseQuery<SearchResult> implements SearchResultQ
...
+      flatten.ad_url_string like ${characterToSearch}
+      or flatten.non_ads_url_string like ${characterToSearch}
+      group by flatten.id, flatten.keyword, flatten.nod_ad_urls, flatten.ad_word_top_urls, flatten.status`
```
olivierobert, line 194: Shouldn't the results be paginated too? (using LIMIT)

src/queries/SearchResultQuery.ts
```diff
@@ -26,42 +32,194 @@ class SearchResultQuery extends BaseQuery<SearchResult> implements SearchResultQ
...
+       AND array_to_string(sr.ad_word_top_urls, ', ') ~ ${queryInput}
+       LIMIT ${this.paginationQuery.take} OFFSET ${this.paginationQuery.skip}`
+    );
+  }
+
+  private async filterByPartialUrl(
+    queryInput: string,
+    leastCharacterCount: number
+  ): Promise<SearchResult[]> {
+    const characterToSearch = `%${queryInput}`.repeat(leastCharacterCount).concat('%');
```
olivierobert, line 176: `leastCharacterCount` is to count the number of occurrences of that character (e.g., search for strings that have `>` at least `3` time, then the string `menu > store > categories > accessories` would be a match) but this would search for `>>>` if `leastCharacterCount` was set to `3.

olivierobert (changes requested):
I recommend dropping the occurrences count to lower the complexity and get this merged

src/queries/SearchResultQuery.ts
Prabeshpd: We are displaying status in table. That's why it is needed 🙂

src/queries/SearchResultQuery.ts
Prabeshpd: I have dropped the filter params. https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/101/commits/89fdc82137977028981366e5b8df793bc65b19be

tednguyendev: The act of passing these params(`queryInput, type, scope`) through many functions seems unnecessary too. As we are in a class and we already have `this.filters`, we just need to get these params from the `this.filters` when we need.

But since that's minor, we can skip it for now ✋

tednguyendev (approved):

olivierobert (approved):

PR #103, opened 2023-09-04
[#105] [UI] As a user, I can filter search results

### LINKED ISSUE #105: [UI] As a user, I can filter search results
## Why

User may want to filter the search results based on different criteria. In requirement it is required to filter for different reasons.

## Acceptance criteria

- Use can filter certain url
- User can filter certain words in ad url
- User can filter whether url contains specific character for n times.

## Resource

- Create a modal to add filters.
- On addition of filters it should call action to filter items

Prabeshpd:
- Close #105 

## What happened 👀

- Added a Filter form

- Add test for filter form component

## Insight 📝

Since, user can filter the search results using various params. A UI element was needed from where to apply the filter. I have used a filter form with a input search. To discern the scope and type of search a dropdown is used.

## Proof Of Work 📹

https://github.com/nimblehq/ic-nextjs-prabesh-dahal/assets/39412015/f19b4f14-dc3f-4550-9edb-eeec3efe136e


src/constants/constants.ts
```diff
@@ -2,3 +2,17 @@ export const DEBOUNCE_DEFAULT_VALUE = 250;
 
 export const PAGINATION_CURRENT_PAGE = 1;
 export const PAGINATION_LIMIT = 10;
+
+export const DEFAULT_BACKGROUND_COLOR = '#111827';
+export const DEFAULT_TEXT_COLOR = '#ffffff';
```
tednguyendev, line 7: Since they are only being used once, let just put them in `MODAL_DEFAULT_STYLES` 🙈

src/constants/constants.ts
```diff
@@ -2,3 +2,17 @@ export const DEBOUNCE_DEFAULT_VALUE = 250;
...
+
+export const MODAL_DEFAULT_STYLES = {
```
tednguyendev, line 9: Do people usually put constants of the FE and BE in the same file? 🤔 Thank you for your explanation in advance 🙏

src/stylesheets/base/_modal.scss
```diff
@@ -0,0 +1,31 @@
...
+    padding: 1rem;
+    border-bottom-width: 1px;
+    border-color: $border-color;
+  }
+
+  &__head > h3 {
+    font-weight: 600;
+    font-size: 1.25rem;
+    line-height: 1.75rem;
+  }
```
tednguyendev, line 15: Will this work? 👀 

```suggestion
  &__head {
    display: flex;
    justify-content: space-between;
    align-items: flex-start;
    padding: 1rem;
    border-bottom-width: 1px;
    border-color: $border-color;

    & > h3 {
      font-weight: 600;
      font-size: 1.25rem;
      line-height: 1.75rem;
    }
  }
```

src/components/SearchResultList/SearchResultModalForm.test.tsx
```diff
@@ -0,0 +1,96 @@
+import * as React from 'react';
+
+import { render, waitFor, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+
+import SearchResultModalForm from './SearchResultModalForm';
+
+describe('search result modal form', () => {
```
tednguyendev, line 8: How about this 💪 

```js
describe('Search result modal form', () => {
  describe('given isOpen is true', () => {
    it('renders the component', async () => {
    });
  });

  describe('given isOpen is false', () => {
    it('does NOT render the component', async () => {
    });
  });

  describe('given form is filled', () => {
    it("submits the form's input values on submit", async () => {
    });
  });
});

```

src/components/SearchResultList/SearchResultList.tsx
tednguyendev, line 1: We may also need to update the test for this file since this PR introduces many new things.

src/constants/constants.ts
Prabeshpd: The projects I have worked on normally have separate folders for frontend or backend or separate repositories. Since, here we have the same shared folders I used the same. This is kind of new for me too. 😅

src/stylesheets/base/_modal.scss
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/103/commits/621c304c6bc944c554cb40a9a211c23728dfb2f0

src/components/SearchResultList/SearchResultList.tsx
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/103/commits/6e920b33d358c5824664e9b4166bd54302a5c927

src/components/SearchResultList/SearchResultModalForm.tsx
```diff
@@ -0,0 +1,149 @@
...
+  const { closeModal, isOpenModal, onApply } = props;
+  return (
+    <div data-testid="search-result-filter-modal">
+      <Modal
+        isOpen={isOpenModal}
+        onRequestClose={closeModal}
+        style={MODAL_DEFAULT_STYLES}
+        ariaHideApp={false}
+      >
+        <div className="ReactModalPortal__head">
```
olivierobert, line 29: Why is the class name casing different? And not:

```suggestion
        <div className="search-result-filter-modal__head">
```

The modal component (https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/103/files#diff-a128ba64d2993a021e074681d30d6e58fcdee69d7054604d91d63d38309ce520) should not use `ReactModalPortal` as the block element (as this comes from the library). Think of combining class names:

```react
<div data-testid="modal search-result-filter-modal">
   <Modal
        isOpen={isOpenModal}
        onRequestClose={closeModal}
        style={MODAL_DEFAULT_STYLES}
        ariaHideApp={false}
      >

  <div className="modal__head">

  #...
</div>
```

`modal` is the shared component that can be used in multiple page, while `search-result-filter-modal` can be used to target this very specific modal

olivierobert (changes requested):

src/components/SearchResultList/SearchResultModalForm.tsx
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/103/commits/e4d939d6e32197ea100fa7403e6fcf3f6d019cb6. Remove the modal using new implementaiton.

src/components/SearchResultList/FilterForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+        onResetFilter();
+      }}
+      onSubmit={async (values) => {
+        const payload = {
+          queryInput: values.queryInput.trim(),
+          scope: values.searchScope?.value,
+          type: values.searchType?.value,
+        };
+
+        await onApplyFilter(payload);
```
tednguyendev, line 46: Seems like it is not needed?

```suggestion
        onApplyFilter(payload);
```

CleanShot 2023-09-12 at 13 01 30@2x

src/stylesheets/base/_react-select.scss
```diff
@@ -0,0 +1,23 @@
+.react-select {
+  margin: 1rem;
+  min-width: 10rem;
+  background-color: map-get($color-black, 'secondary');
+  color: map-get($color-white, 'primary');
```
tednguyendev, line 5: Let's make it looks a bit more interactive

```suggestion
  color: map-get($color-white, 'primary');
  cursor: pointer;
```

CleanShot 2023-09-12 at 13 03 17@2x

tednguyendev (approved):

src/components/SearchResultList/FilterForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { SearchTypeOptions, SearchScopeOptions } from '@/constants/constants';
+import { SearchScope, SearchType } from '@/constants/enum';
+import { SearchResultFilter } from '@/types/query';
+
+interface InjectedProps {
+  onApplyFilter: (filterParams: SearchResultFilter) => void;
+  onResetFilter: () => void;
+}
+
+interface FormValues {
```
olivierobert, line 17: Same idea about naming:

```suggestion
interface FilterFormValues {
```

olivierobert (commented):
Mostly minor issues that needs to be tightened up

src/stylesheets/base/_react-select.scss
```diff
@@ -0,0 +1,23 @@
...
+
+  & &__control {
```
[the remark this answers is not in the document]
Prabeshpd: React select needs to pass the extender class and can be only be accessed with both of the attributes meaning

```<ReactSelect className="react-select" extendedClass="react-select" />```

And for the scss

```.react-select .react-select__control```

For avoiding too many names I used the same string for class name and extended class name. Does it make sense to add another name for extended ?
new way 
```<ReactSelect className="react-select" extendedClass="react-select-container" />```


```.react-select .react-select-container__control```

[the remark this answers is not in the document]
tednguyendev: By `extendedClass` you mean the prop [`classNamePrefix`](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/blob/529851b4d8afe27bb894e1cb33430f5c31478d5a/src/components/SearchResultList/FilterForm.tsx#L73) right? 👀 

Ah, so by arranging the CSS selectors(`& &__control`) like that, you increase the [CSS specificity](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/103/files) to override the style.

Yes having the comment is better 👍

olivierobert (approved):

PR #117, opened 2023-09-07
[Chore] [#87] Move the test to source folders

### LINKED ISSUE #87: [Chore] Move all the test to source folder
## Why

Current test resides in two places. Make sure to move the test to src folders so it follows javascript conventions

## Acceptance Criteria

- Move all the  tests to source folder file to interact with database.

Prabeshpd:
- Close #87 

## What happened 👀

- Moved the test to src folders.
- Change the factory object to function

## Insight 📝
Since, all the test should be in source folders. Moved all the test to source folders.  We were using two projects to run the file in different folders either to mock the database or to use a real database operations.  Since, we remove the test which mock database. `<rootDir>/jest.unit.config.js` project is not needed anymore.

## Proof Of Work 📹

Passed CI pipelines.


package.json
```diff
@@ -17,7 +17,6 @@
   },
   "jest": {
     "projects": [
-      "<rootDir>/jest.unit.config.js",
```
olivierobert, line 20: Why was this removed? It is best to document this kind of decision in the PR " Insight " section.

src/app/actions/upload.test.ts
```diff
@@ -4,14 +4,15 @@
 
 import fs from 'fs';
 
-import { uploadCsv } from '../../src/app/actions/upload';
-import { user } from '../factories/user';
-
 import dbClient from '@/config/database';
 
+import { uploadCsv } from './upload';
+import { userFactory } from '../../../test/factories/user';
```
olivierobert, line 10: How about setting an alias (`@test`) to avoid typing `../../../`

olivierobert (commented):
Some tests might still make use of a DB mock. It is not an all-or-nothing strategy. For now, since all tests do DB inserts, it is fine to remove `'<rootDir>/test/database.ts'` but it could be brought in again if needed.

package.json
Prabeshpd: We were using two projects to run the file in different folders either to mock the database or to use a real database operations.  Since, we remove the test which mock database. It is not needed anymore.

olivierobert (approved):

tednguyendev (approved):

PR #129, opened 2023-09-11
[#88] [API] As a user, I can sign in

### LINKED ISSUE #88: [API] As a user, I can sign in
## Why

There is no API implementation for sign in using credentials or Google OAuth strategy. So, an api endpoint must be made

## Acceptance Criteria

- Implement an API endpoint to authenticate a user
- Ensure all API endpoints require to be authenticated

## Resources 
- Use a database adapter the next-auth provides to manage schema and store tokens
- Use database adapters [Adapters](https://authjs.dev/reference/core/adapters)
- [Primsa Native Support](https://authjs.dev/reference/adapter/prisma)

Prabeshpd:
- Close #88 

## What happened 👀

- Added prisma adapter for next-auth
- Removed the unused functions

## Insight 📝

- Since, we are using next-auth native feature for oauth login. Schemas are changed as per defined in the documentation. Next-auth will add the access_token on the database storage for the google login.
- For credentials it does not add those changes to database but rather manages them internally.

## Proof Of Work 📹

https://github.com/nimblehq/ic-nextjs-prabesh-dahal/assets/39412015/2157272b-f15f-48c4-b1e1-d77cd2108a2d

https://github.com/nimblehq/ic-nextjs-prabesh-dahal/assets/39412015/c0e4bad1-23a9-4fd9-b0da-0b3c35d1ab01

Screenshot 2566-09-11 at 17 39 13


prisma/migrations/20230911095147_users/migration.sql
```diff
@@ -2,8 +2,10 @@
 CREATE TABLE "users" (
     "id" SERIAL NOT NULL,
     "name" VARCHAR(100) NOT NULL,
-    "email" VARCHAR(50) NOT NULL,
-    "password" VARCHAR(100) NOT NULL,
+    "email" VARCHAR(50),
```
olivierobert, line 5: That could be still NOT NULL, as in both auth flows, the email is always provided, no?

src/services/auth/auth.ts
```diff
@@ -20,7 +11,7 @@ export async function authenticateCredentialLogin(credentials: Credentials) {
 
   const { email, password } = credentials;
   const user = await findByEmail(email);
-  if (!user) {
+  if (!user || !user.password) {
     throw new AuthError({ message: 'Provided email or password does not match.' });
   }
```
olivierobert, line 16: Could that conditional be removed or reworked as I am not following why the additional `!user.password` was added 🤔 I would expect a single condition:

```js
if (!user || !isMatchedPassword) {
    throw new AuthError({ message: 'Provided email or password does not match.' });
}
```

Whether there is no user, no password, or the provided password does not match, the same error applies.

prisma/schema.prisma
```diff
@@ -41,18 +44,42 @@ model SearchResult {
...
 }
 
-model Identity {
-  id           String                   @id @default(uuid()) @db.Uuid
-  provider     AuthProviders            
-  user         User                     @relation(fields: [userId], references: [id])
-  userId       Int                      @db.Integer
-  createdAt    DateTime @default(now()) @map("created_at")
-  updatedAt    DateTime @default(now()) @updatedAt @map("updated_at")
+model Account {
```
olivierobert, line 47: Is there a record populated in that table for a user authentication with email? Will the provide be `credentials`?

prisma/schema.prisma
```diff
@@ -8,14 +8,17 @@ datasource db {
...
-  createdAt    DateTime       @default(now()) @map("created_at")
-  updatedAt    DateTime       @default(now()) @updatedAt @map("updated_at")
-  SearchResult SearchResult[]
-  Identity     Identity[]
+  id            Int            @id @default(autoincrement())
+  name          String         @db.VarChar(100)
+  email         String?        @db.VarChar(50)
+  emailVerified DateTime?
+  password      String?        @db.VarChar(100)
+  image         String?
```
olivierobert, line 16: Could that be named `profileImage`?

olivierobert (commented):
I have some questions and concerns but overall the current implementation would be sufficient indeed.

prisma/migrations/20230911095147_users/migration.sql
Prabeshpd: Yes, will update it. Had done rough implementation first just to confirm things. 🙂.

prisma/schema.prisma
Prabeshpd: Would need to try it. Will try it by adding @map

src/services/auth/auth.ts
```diff
@@ -20,46 +19,10 @@ export async function authenticateCredentialLogin(credentials: Credentials) {
 
   const { email, password } = credentials;
   const user = await findByEmail(email);
-  if (!user) {
-    throw new AuthError({ message: 'Provided email or password does not match.' });
-  }
-
-  const isMatchedPassword = await crypt.compareWithHashValue(password, user.password);
+  const isMatchedPassword = await hasUserPasswordMatched(user, password);
```
olivierobert, line 22: On eon side the variable name uses `is` and on the other the method uses `has`, so it is kind of inconsistent. How about the following:

```suggestion
  const isValidPassword = await validateUserPassword(user, password);
```

Naming things is hard 💦

src/services/auth/auth.ts
```diff
@@ -1,17 +1,16 @@
...
 import { Credentials } from '@/types/common';
 
 import AuthError from './error';
-import { createUserWithIdentity, mergeExistingUser } from '../user/user';
 
-interface Profile {
-  email?: string;
-  name?: string;
-}
+export const hasUserPasswordMatched = async (user: User | null, password: string) => {
```
olivierobert, line 9: Since this method will not be used elsewhere, the need to export it is minimal. I understand it was likely done to make it testable, but the functionality can be already tested in `authenticateCredentialLogin`. If you still prefer to have tests, I recommend moving it to `lib/auth.ts` (this could reworked from the existing `lib/crypt.ts` ti have everything in one file).

Prabeshpd: Totally makes sense. I will remove the export part here. Since, it is already tested in authenticate Credential. Will remove the test too. Will make that more simpler.

src/services/auth/auth.test.ts
```diff
@@ -19,14 +17,16 @@ describe('Auth:', () => {
         const userPayload = { name: user.name, email: user.email, password: user.password };
         await createUser(userPayload);
 
+        const email = user.email || '';
+        const password = user.password || '';
```
tednguyendev, line 21:
```suggestion
        const userPayload = { name: user.name, email: user.email, password: 'password' };
        await createUser(userPayload);

        const email = user.email;
        const password = 'password';
```

Better to do the `|| ''` to bypass typescript, I think 🙈

src/services/user/user.ts
```diff
@@ -1,19 +1,17 @@
...
 
-interface UserCreateIdentityInput {
-  user: { email: string; name: string };
-  authProvider: AuthProviders;
-}
-
 export async function createUser(payload: Prisma.UserCreateInput) {
+  if (!payload.email || !payload.password) {
+    throw new UserError({
+      message: 'Email and Password credentials must be provided to create an account.',
```
tednguyendev, line 11: Or just:

```suggestion
      message: 'Please enter email and password.',
```

tednguyendev (approved):
Some minors but LGTM 👍

olivierobert (approved):

PR #130, opened 2023-09-12
[#110] [Backend] Implement scraper for Bing

### LINKED ISSUE #110: [Backend] Implement scraper for Bing
## Why

In addition to Google, users want to be able to get scraped data from the other major search engines.

## Acceptance Criteria

- By default, each uploaded keyword is scaped on both Google and Bing. So it creates two search results records for each keyword.
- The scraping must be handled asynchronously.

Prabeshpd:
- Close #110 

## What happened 👀

- Add scraper for bing
- Add parser for bing
- Update scraper job to add job for bing
- Add column search engine with enum.

## Insight 📝

- Implemented scraper for bing. Added a job for bing to scrape when the file are uploaded. Have created an enum for database to store which engine searched the result. 
- In terms of parser have created a separate class with same interface as google parser. I did not create a separate parent class as the implementation is different and it is good to have a interface definition.

## Proof Of Work 📹

Passed CI pipelines for test


src/services/parser/bingParser.test.ts
```diff
@@ -0,0 +1,179 @@
+/**
+ * @jest-environment node
+ */
+
+import BingParser from './bingParser';
+
+describe('BingParse:', () => {
```
olivierobert, line 7: SHouldnt this be:

```suggestion
describe('BingParser:', () => {
```

src/services/parser/bingParser.test.ts
```diff
@@ -0,0 +1,179 @@
...
+  describe('given valid raw html for bing', () => {
+    describe('parseTopAdWordList', () => {
+      describe('given raw html with top ad words', () => {
```
olivierobert, line 10: Any way way to limit the nesting here as it looks like [the pyramid of doom](https://en.wikipedia.org/wiki/Pyramid_of_doom_(programming)#)

src/services/parser/parser.ts
```diff
@@ -0,0 +1,29 @@
...
+
+import BingParser from './bingParser';
+import ParseError from './error';
+import GoogleParser from './googleParser';
+
+class Parser {
+  searchEngine: SearchEngine;
+  htmlRawString: string;
+
+  constructor(searchEngine: SearchEngine, htmlRawString: string) {
```
olivierobert, line 11: It would be best to raise an exception at the initialization phase:

```js
constructor(searchEngine: SearchEngine, htmlRawString: string) {
  if (!isSupportedSearchEngine(SearchEngine)) {
    throw new ParseError({ message: `Unsupported search engine: ${this.searchEngine}` });
  }

  this.searchEngine = searchEngine;
  this.htmlRawString = htmlRawString;

  return // set the parser
}

private isSupportedSearchEngine() {
  return SEARCH_ENGINES.incldues(this.searchEngine)
}
``` 

The earlier it fails the better. It would also follow the factory pattern.

src/config/parserSelectorConfig.ts
```diff
@@ -0,0 +1,32 @@
...
+    },
+    bingSelectors: {
+      adLinkElementAttribute: process.env.AD_LINK_ANCHOR_ELEMENT_ATTRIBUTE_BING || '.b_adurl',
+      adTopParent: process.env.AD_TOP_PARENT_BING || '.b_ad, .b_adTop',
+      adWords: process.env.AD_WORDS_BING || '.sb_ad, .sb_adTA',
+      linkElement: process.env.LINK_ELEMENT_BING || 'a',
+      adBottomParent: process.env.AD_BOTTOM_PARENT_BING || '.b_ad, .b_adBottom',
+      nonAdWords: process.env.NON_AD_WORDS_BING || '.b_algo',
+      linkElementAttribute: process.env.LINK_ELEMENT_ATTRIBUTE_BING || 'href',
+      nonAdParent: process.env.NON_AD_PARENT_BING || '#b_results',
```
olivierobert, line 27: It might be better to leave this defined in each parse classes. these are not reused apart from tests and would help in understanding each parser better.

I also recommend dropping the need to use env variables as this does not work anymore when having more than 1 search engine. The benefits outweigh the cons. So let's keep these defined as contacts in the parser files 💡

src/services/parser/bingParser.ts
```diff
@@ -0,0 +1,82 @@
...
+
+import parserSelectorConfig from '@/config/parserSelectorConfig';
+import { ISelectorId } from '@/types/common';
+import { IParse } from '@/types/parse';
+
+const {
+  searchEngineSelectorIds: { bingSelectors },
+} = parserSelectorConfig;
+
+class BingParser implements IParse<cheerio.CheerioAPI> {
```
olivierobert, line 11: How about defining a base parser that each parser muct implement. This way `IParse<cheerio.CheerioAPI>` would be defined only once (a win) and we could ensure that all parsets expose the same API.

olivierobert (commented):
Some early feedback. More will follow.

src/services/parser/bingParser.ts
Prabeshpd: I intentionally did not create the base class as the internal implementation are different. I wanted to bind class with interface rather than extending  it. But I think it makes sense to create a base class. Will create it.

src/services/parser/bingParser.test.ts
Prabeshpd: I think we can remove the valid html part here as the empty string already accounts for that.

src/app/actions/upload.ts
```diff
@@ -29,9 +30,19 @@ export async function uploadCsv(file: FormDataEntryValue | null, userId: number)
       throw error;
     }
 
-    const searchResultPayload = keywords.map((keyword) => ({ keyword: keyword, userId: +userId }));
+    const payload = keywords.reduce(
+      (accumulator: Prisma.SearchResultCreateManyInput[], keyword) => {
+        accumulator.push(
+          { keyword, userId: +userId, searchEngine: SearchEngine.bing },
+          { keyword, userId: +userId, searchEngine: SearchEngine.bing }
```
olivierobert, line 37: One of them should be Google, right? This would enqueue later the scarping for Bing twice. 

```suggestion
          { keyword, userId: +userId, searchEngine: SearchEngine.google }
```

src/app/actions/upload.ts
```diff
@@ -29,9 +30,19 @@ export async function uploadCsv(file: FormDataEntryValue | null, userId: number)
       throw error;
     }
 
-    const searchResultPayload = keywords.map((keyword) => ({ keyword: keyword, userId: +userId }));
+    const payload = keywords.reduce(
+      (accumulator: Prisma.SearchResultCreateManyInput[], keyword) => {
```
olivierobert, line 34: It would be best to avoid `accumulator` as this is from the spec 🙈 Also, here [flatMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap) would be sufficient:

```js
const payload = keywords. flatMap((keyword) => {
  return [
     { keyword, userId: +userId, searchEngine: SearchEngine.bind },
     { keyword, userId: +userId, searchEngine: SearchEngine.google },
  ]
})
```

src/services/parser/baseParser.ts
```diff
@@ -0,0 +1,37 @@
...
+class BaseParser implements IParse<cheerio.CheerioAPI> {
+  htmlRawString: string;
+  htmlDocumentAPI: cheerio.CheerioAPI;
+
+  constructor(htmlRawString: string) {
+    this.htmlRawString = htmlRawString;
+    this.htmlDocumentAPI = this.loadHtmlDocumentApi();
+  }
+
+  private loadHtmlDocumentApi() {
```
olivierobert, line 14: It would be best to move all private methods below all public ones for readability

src/services/parser/baseParser.ts
```diff
@@ -0,0 +1,37 @@
...
+    return cheerio.load(this.htmlRawString);
+  }
+
+  parseTopAdWordList() {
+    const urls: string[] = [];
```
olivierobert, line 19: Since a base class is not meant to be used directly (like an abstract class), the methods should be defined but raise a `NotImplementException` (see how Nest.JS does it: http://nestjs-doc.exceptionfound.com/classes/notimplementedexception.html)

src/workers/scraper.ts
```diff
@@ -1,32 +1,34 @@
...
   try {
-    const htmlData = await searchResult(keyword);
-    const parserInstance = new Parser(selectorIdsSearch, htmlData);
-    const topAdParsed = parserInstance.parseTopAdWordList();
-    const bottomAdParsed = parserInstance.parseBodyAdWordList();
-    const nonAdParsed = parserInstance.parseNonAdWordList();
-    const adWordsCount = topAdParsed.count + bottomAdParsed.count;
-    const nonAdUrls = nonAdParsed.urls;
+    const htmlData = await searchResult(keyword, searchEngine);
+    const parserInstance = new Parser(searchEngine, htmlData).getParserInstance();
```
olivierobert, line 12: When using `instance`, the first thought is that we use a singleton pattern, but that is not what the code does here. So, I recommend not using that term.

It can be simplified to:

```suggestion
    const parser = new Parser(searchEngine, htmlData).getParserInstance();
```

Technically, this would also work:

```suggestion
    const parser = new Parser(searchEngine, htmlData).parser;
```

So an extra method is not needed 💡

src/workers/scraper.ts
```diff
@@ -1,32 +1,34 @@
...
+    const topAdUrls = parserInstance.parseTopAdWordList();
+    const bottomAdUrls = parserInstance.parseBottomAdWordList();
+    const nonAdUrls = parserInstance.parseNonAdWordList();
+    const adWordsCount = topAdUrls.length + bottomAdUrls.length;
+    const totalLinksCount = adWordsCount + nonAdUrls.length;
```
olivierobert, line 17: A line break would be nice for readability

```suggestion
    const totalLinksCount = adWordsCount + nonAdUrls.length;
    
```

test/fixtures/bingParse.html
olivierobert, line 1: Naming-wise, how about: `bing-search-result.html`?

src/workers/queueScraper.ts
```diff
@@ -10,12 +10,16 @@ const BASE_RETRY_DELAY_MS = 1000;
 const queueScrapeJob = async (job: Job) => {
   const { userId } = job.data;
 
-  const searchResults = await listSearchResultByStatus(userId);
+  const searchResults = await listSearchResultByStatus(+userId);
```
olivierobert, line 13: Why do we need the plus sign: **+**userId?

src/services/search/searchResult.test.ts
```diff
@@ -0,0 +1,65 @@
...
+    describe('given search engine is INVALID', () => {
+      it('throws error', async () => {
+        await expect(searchResult('node', 'explorer')).rejects.toThrowError(
+          'Unsupported search engine: explorer'
+        );
+      });
+    });
+  });
+
+  describe('given axios call REJECTS with error', () => {
```
olivierobert, line 44: Ditto here:

```suggestion
  describe('given a failed request to fetch the search result page', () => {
```

src/services/search/searchResult.test.ts
```diff
@@ -0,0 +1,65 @@
...
+
+import request from '@/lib/searchRequest';
+
+import { searchResult } from './searchResult';
+
+jest.mock('../../lib/searchRequest.ts');
+const mockedAxios = request as jest.Mocked<typeof request>;
+
+describe('searchResult', () => {
+  describe('given axios call resolves data', () => {
```
olivierobert, line 14: Axios could be left out of the test description:

```suggestion
  describe('given a succesful request to fetch the search result page', () => {
```

olivierobert (changes requested):

src/workers/queueScraper.ts
Prabeshpd: It internally maps the data to string and we may get a error. I faced a similar situation while testing. So, converting it to integer.

src/config/config.ts
```diff
@@ -36,16 +33,6 @@ const config: Configuration = {
     secret: process.env.NEXT_AUTH_SECRET || 'ENTER_NEXT_AUTH_SECRET',
     saltRounds: '11',
   },
-  selectorIdsSearch: {
-    adLinkElementAttribute: process.env.AD_LINK_ANCHOR_ELEMENT_ATTRIBUTE || '.sVXRqc',
```
tednguyendev, line 40: Let's also remove these from `.env.example`

src/services/parser/bingParser.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  parseTopAdWordList() {
+    const urls: string[] = [];
+    const adElementsDom = this.htmlDocumentAPI(this.selectorIds.adTopParent);
+
+    const adUrlElements = this.htmlDocumentAPI(adElementsDom).find(
+      `${this.selectorIds.adTopParent} ${this.selectorIds.adLinkElementAttribute}`
+    );
+
+    adUrlElements.each((index, element) => {
```
tednguyendev, line 32: As we not use `index`

```suggestion
    adUrlElements.each((_, element) => {
```

src/services/parser/parser.ts
```diff
@@ -0,0 +1,30 @@
...
+      throw new ParseError({ message: `Unsupported search engine: ${searchEngine}` });
+    }
+
+    switch (searchEngine) {
+      case SearchEngine.bing:
+        this.parser = new BingParser(htmlRawString);
+        break;
+      case SearchEngine.google:
+        this.parser = new GoogleParser(htmlRawString);
+        break;
```
tednguyendev, line 21: 👀 

```suggestion
```

src/services/search/searchResult.test.ts
```diff
@@ -0,0 +1,65 @@
...
+      it('returns resolved data', async () => {
+        mockedAxios.get.mockResolvedValue({ data: { message: 'success' } });
+
+        await expect(searchResult('node', SearchEngine.bing)).resolves.toEqual({
+          message: 'success',
+        });
+      });
+    });
+
+    describe('given search engine is INVALID', () => {
```
tednguyendev, line 35: Should this describe `given search engine is INVALID` inside the describe `given a successful request to fetch the search result page` 🙈 

How abt:

```ts
describe('searchResult', () => {
  describe('given search engine is "google"', () => {
    describe('given search successfully', () => {
      it('returns resolved data', async () => {
      });
    });
    describe('given search FAILED', () => {
      it('throws an error', async () => {
      });
    });
  });

  describe('given search engine is "bing"', () => {
    describe('given search successfully', () => {
      it('returns resolved data', async () => {
      });
    });
    describe('given search FAILED', () => {
      it('throws an error', async () => {
      });
    });
  });

  describe('given search engine is INVALID', () => {
    it('throws error', async () => {
    });
  });
});

```

olivierobert (approved):
Looks good 👍

tednguyendev (commented):
LGTM 🤌

tednguyendev:
^Just realized I commented instead of approved 🤣

PR #131, opened 2023-09-12
[#118] [UI] As a user, I can view scraped information from Bing

### LINKED ISSUE #118: [UI] As a user, I can view scraped information from Bing
## Why

Once the search result page has been scraped, a user wants to see it.

## Acceptance Criteria

- In the search results list, add a column to identify each row as coming from Google or Bing
- Add a new page to view the craped information from Bing for a keyword

Prabeshpd:
- Close #118 

## What happened 👀

- Added the column of search engine
- Added search engine column in query
- Added search engine in detail page

## Insight 📝

Since we have added a new column for database to know which search engine is used. For viewing a table column is added and it is added on the detail page. Also the query is updated to include it

## Proof Of Work 📹

https://github.com/nimblehq/ic-nextjs-prabesh-dahal/assets/39412015/a610e2af-ca9e-4fa5-b3cb-5a57e46108d4


tednguyendev (approved):

src/components/SearchResultDetail/SearchResultDetail.tsx
```diff
@@ -60,6 +60,9 @@ const SearchResultDetail = ({ id }: { id: string }) => {
           <p className="detail-search-result__info">
             Total number of non ad words: {searchResult?.nonAdLinksCount || 0}
           </p>
+          <p className="detail-search-result__info">
+            Search Engine: {searchResult?.searchEngine || ''}
```
olivierobert, line 64: There is no case when this can be null, or something is very wrong with the application. So in the latter case, it is best to let the application crash/error out.

src/components/SearchResultList/SearchResultList.test.tsx
```diff
@@ -32,6 +32,7 @@ describe('Lists search result', () => {
 
       expect(screen.queryByText('Keyword')).toBeInTheDocument();
       expect(screen.queryByTestId('pagination')).toBeNull();
+      expect(screen.getByTestId('table-header-columns').children.length).toEqual(4);
```
olivierobert, line 35: That is a detail of implementation that can change (say the UI changes). It would be best to check each item in the list has the expected terms on the page

olivierobert (approved):
Just some minor improvement suggestions but none are blockers. So I leave it to you to decide to adopt them or not.

src/components/SearchResultList/SearchResultList.test.tsx
Prabeshpd: Updated the Test https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/131/commits/e5a6255c50b1fbefe978a92ab99aa29b7d80ebab

PR #132, opened 2023-09-13
[#112] [Backend] As user, I can export the HTML for a search result to a PDF

### LINKED ISSUE #112: [Backend] As user, I can export the HTML for a search result to a PDF
## Why

In addition to viewing the scraped page on the application, a user should be able to download it for safekeeping.

## Acceptance Criteria

- Add a new server action to generate a PDF from the HTML of a search result.
- pending on how long it takes, it can be processed asynchronously or  asynrchnosnouly.

Prabeshpd:
- Close #112 

## What happened 👀

- Created a function for converting html to pdf using puppeteer
- Created a server action to return pdf base 64 value to client
- Add tests

## Insight 📝
Used puppeteer to convert the html to pdf. The approach goes with conversion of html to pdf on request since, the conversion files are not large. Puppeteer converts the html and the server action converts it to base 64 string to present it to front end.

>[!Note]
> - I have not tested the html to pdf as it is implementation of a third party library. 
> -  For testing export function I have mocked the response of a pdf buffer and ensured my business logic is working fine.

## Proof Of Work 📹

Passed CI pipelines of test


src/lib/file.ts
```diff
@@ -22,3 +24,16 @@ export async function parseCSVFormData(file: FormDataEntryValue): Promise<string
 
   return bufferString.split(',\n').map((data) => data.trim());
 }
+
+export async function convertHtmlToPdf(htmlString: string) {
```
olivierobert, line 28: This is more than a helper and it relies on a third-party library, so what do you think of moving it to `src/services/export/pdf.js`?

src/app/actions/exportPdf.test.ts
```diff
@@ -0,0 +1,80 @@
...
+  const serverSessionUtilities = { getServerSession };
+  const fileUtilities = { convertHtmlToPdf };
+  const mockedServerSession = serverSessionUtilities as jest.Mocked<{
+    getServerSession: () => Promise<SessionWithId | null>;
+  }>;
+  const mockedFile = fileUtilities as jest.Mocked<{
+    convertHtmlToPdf: () => Promise<Buffer>;
+  }>;
+
+  describe('given user is authorized', () => {
```
olivierobert, line 30: It should be:

```suggestion
  describe('given user is authenticated', () => {
```

Authorization would be about checking user roles and permissions. There is no such checks in place now 🙈

olivierobert (commented):
The code is clear and simple 👍 So no major concerns parts from where to place the code

src/app/actions/exportPdf.test.ts
```diff
@@ -0,0 +1,80 @@
...
+        const createdUser = await createUser(user);
+        const htmlString = '<div>Test Html</div>';
+        const pdfBuffer = await fs.readFileSync('./test/fixtures/sample-pdf.pdf');
+        const searchResultPayload = { ...searchResult, html: htmlString, userId: createdUser.id };
+        await createManySearchResults([searchResultPayload]);
+        mockedServerSession.getServerSession.mockResolvedValue({
+          user: { id: createdUser.id.toString() },
+          expires: '',
+        });
+        mockedFile.convertHtmlToPdf.mockResolvedValue(pdfBuffer);
```
tednguyendev, line 44:
```suggestion
        const user = userFactory();
        const createdUser = await createUser(user);
        const searchResult = searchResultFactory();
        const htmlString = '<div>Test Html</div>';
        const searchResultPayload = { ...searchResult, html: htmlString, userId: createdUser.id };
        await createManySearchResults([searchResultPayload]);

        mockedServerSession.getServerSession.mockResolvedValue({
          user: { id: createdUser.id.toString() },
          expires: '',
        });

        const pdfBuffer = fs.readFileSync('./test/fixtures/sample-pdf.pdf');
        mockedFile.convertHtmlToPdf.mockResolvedValue(pdfBuffer);
```

src/app/actions/exportPdf.test.ts
```diff
@@ -0,0 +1,80 @@
...
+        expect(pdf).not.toBeNull();
+        expect(pdf.includes(mimeTypeIdentifierPDF)).toEqual(true);
+      });
+    });
+
+    describe('given search result with NO html', () => {
+      it('rejects with 422 error', async () => {
+        const user = userFactory();
+        const searchResult = searchResultFactory();
+        const createdUser = await createUser(user);
```
tednguyendev, line 57:
```suggestion
        const user = userFactory();
        const createdUser = await createUser(user);
        const searchResult = searchResultFactory();
```

tednguyendev (approved):
It look neat ✋

I also agree with the approach to mock the response of the pdf.

👇 Just some minor suggestions to rearrange the code a bit for better readability. And remember to fix the lint too.

olivierobert (approved):

PR #135, opened 2023-09-17
[#120] [Chore] Set up Cypress for UI testing

### LINKED ISSUE #120: Set up Cypress for UI testing
## Why

To implement UI testing, Cypress must be set up.

## Acceptance Criteria

- Add Cypress as a dependency
- Configure Cypress to run locally 
- Add a new step or job to run UI tests on GitHub Actions

Prabeshpd:
- Close #120 

## What happened 👀

- Added cypress for testing
- Added github actions to run cypress on CI pipelines
- Added test for user sign up.
- Added cypress task to seed user and reset them.

## Insight 📝

- Added cypress for UI testing. 
- Added a separate github action job as the existing test is taking a lot of time to run. Since, the cypress test is run parallel it helps in time reduction. Along with that it segregates the UI testing.
- For database seed I have allowed user to pass custom parameters so that one can have more control in testing.
- I have not created a single file of seed as developer will not have control over seeding at testing.
- Along with that it becomes difficult for other to validate the test just looking at the tests as the data validated are seeded separately.

## Proof Of Work 📹

Passed CI pipelines new job for cypress.


.github/workflows/test.yml
```diff
@@ -55,3 +55,27 @@ jobs:
 
       - name: Test
         run: yarn test
+
+  test-ui:
+    name: Test UI
+    runs-on: ubuntu-latest
```
olivierobert, line 61: UI tests can get stuck sometimes. so it is best to add a timeout:

```suggestion
    timeout-minutes: 10
    runs-on: ubuntu-latest
```

.github/workflows/test.yml
```diff
@@ -55,3 +55,27 @@ jobs:
...
+    steps:
```
olivierobert, line 62: A DB will also be required, right?

Prabeshpd: No, I am thinking of intercepting the request to server actions and mocking the response from fixtures. Since, we are throughly checking the database interaction in backend. I was thinking checking the interaction of UI in success and errors.  

I am preparing a document of my understandings and the approaches to get verified.

olivierobert: UI tests must be as "real" as possible, i.e., they must be used to test the application exactly how the user would use it (especially when using Cypress). So, mocking is not expected.

Prabeshpd: Fixed here https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/135/commits/3c01c1a630a982c54f4a7ccdba6a1b00d5bd35b7

.github/workflows/test.yml
```diff
@@ -55,3 +55,53 @@ jobs:
...
+
+      - name: Set Up Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: '18'
+
+      - name: Install Dependencies
+        run: yarn
+
+      - name: Lint
```
tednguyendev, line 95: Do you think this is redundant as we already tested lint above? ✋

.github/workflows/test.yml
```diff
@@ -55,3 +55,53 @@ jobs:
 
       - name: Test
         run: yarn test
+
+  test-ui:
```
tednguyendev, line 59: Since we use the same `services` and `steps` with the current [`test` job](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/blob/98fd1bcc6100a033252b9a38359e9bc219144ea1/.github/workflows/test.yml#L11), are there anyway we can make this DRY? 👀

.github/workflows/test.yml
tednguyendev: Lets also remove [this](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/135/files#diff-271661623b3a6fdb66b408c573ac77f5c190aa725a30bd9a79cefecfa86be5ab)

test/helpers/database.reset.ts
```diff
@@ -0,0 +1,11 @@
+import dbClient from '@/config/database';
+
+const resetDatabase = async () => {
+  return await dbClient.$transaction([
+    dbClient.searchResult.deleteMany(),
+    dbClient.account.deleteMany(),
+    dbClient.user.deleteMany(),
```
tednguyendev, line 7: Another way is to use the [cascade delete](https://www.prisma.io/docs/concepts/components/prisma-schema/relations/referential-actions). But this way works too ✋ 

```suggestion
    dbClient.user.deleteMany(),
```

.github/workflows/test.yml
Prabeshpd: We can define composite actions or reusable-workflow, I guess but  using it would create more hindrances and have their own limitations. I guess we can remove the lint step and use github actions of cypress which reduces the extra code. But I normally add these in each jobs as they run on a separate instance. It helps me to understand the flow more clearly. 

It is a nice context to gather more expert thoughts? cc @olivierobert

olivierobert: There can be an inclination to try to reuse some of the steps, but it is best to keep these as separate as the complexity trade-offs are not worth it.

cypress/e2e/signUp/signUp.cy.ts
```diff
@@ -0,0 +1,53 @@
+const signUpPageTestIds = {
+  inputName: 'sign-up-form__input-name',
+  inputEmail: 'sign-up-form__input-email',
+  inputPassword: 'sign-up-form__input-password',
+  submitButton: 'sign-up-form__submit-button',
```
olivierobert, line 5: The convention is to use`data-set-id=""` and not re-using the BEM CSS class names to keep the test ID simpler:

```
inputName: 'sign-up-form-name',
inputEmail: ''sign-up-form-email',
inputPassword: 'sign-up-form--password',
submitButton: ''sign-up-form-submit',
``` 

See how it is set up: https://github.com/nimblehq/varun-web/blob/e968178cd948bab91ae748264f869564993cd729/src/setupTests.ts#L9, and used: https://github.com/nimblehq/varun-web/blob/develop/cypress/integration/Authentication/index.spec.ts

cypress/support/commands.ts
```diff
@@ -0,0 +1,13 @@
+/// <reference types="cypress" />
+
+import '@testing-library/cypress/add-commands';
+
+Cypress.Commands.add('checkToastMessage', (type: 'success' | 'error', message: string) => {
```
olivierobert, line 5: How about naming it:

```suggestion
Cypress.Commands.add('hasToastMessage', (type: 'success' | 'error', message: string) => {
```

The following reads well:

```
cy.hasToastMessage('error', 'An issue occurred when creating the user account');
```

cypress/support/commands.ts
```diff
@@ -0,0 +1,13 @@
...
+  const toastId = type === 'success' ? 'toast-success' : 'toast-error';
+
+  cy.get(`[id=${toastId}]`).should('contain', message);
+});
+
+Cypress.Commands.add('checkSpinner', (ariaLabel: string) => {
```
olivierobert, line 11: Same suggestion to improve descriptiveness of the command:

```suggestion
Cypress.Commands.add('hasActiveSpinner', (ariaLabel: string) => {
```

cypress.config.ts
```diff
@@ -0,0 +1,26 @@
...
+  e2e: {
+    baseUrl: 'http://localhost:3000',
+    setupNodeEvents(on, config) {
+      on('task', {
+        async reset() {
+          return resetDatabase();
+        },
+
+        async seedUser({ email, password }) {
+          return seedUser(email, password);
```
olivierobert, line 15: The code is quite confusing as it seems to call twice the same method (due to the usage of the same function name), but it is not. Also, the params are not defined anywhere, right? So, can't this be:

```js
async setup({ email, password }) {
   return createUser();
})
```

olivierobert (changes requested):

cypress.config.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/135/commits/c2c9f9d9c25a2a9da36c08f87d148ddcf12fc65e. Have created a event for setup User. Can be changed to setup once the seeding of search result takes place.

cypress/support/e2e.ts
```diff
@@ -0,0 +1,6 @@
+import { configure } from '@testing-library/cypress';
+
+import './commands';
+import './type';
+
+configure({ testIdAttribute: 'data-testid' });
```
olivierobert, line 6: The team coventions is to name it:

```suggestion
configure({ testIdAttribute: 'data-test-id' });
```

As raised in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/135#discussion_r1329552737

cypress/e2e/signUp/signUp.cy.ts
```diff
@@ -0,0 +1,53 @@
...
+        'The user account has been successfully created. You can now log in.'
+      );
+    });
+  });
+
+  describe('given email is NOT unique', () => {
+    it('throws error', () => {
+      const email = 'dev@nimblehq.co';
+      const password = 'password';
+      cy.visit('/auth/signup');
```
tednguyendev, line 39: Let's insert a line break to separate the `arrange` from the `act`(as in `Arrange/Act/Assert (AAA)`) ✋ 

```suggestion
      const password = 'password';
      
      cy.visit('/auth/signup');
```

cypress.config.ts
```diff
@@ -0,0 +1,26 @@
+import { defineConfig } from 'cypress';
+import resetDatabase from '@test/helpers/database.reset';
+import { createUser } from '@test/helpers/database.seed';
+
+export default defineConfig({
+  e2e: {
+    baseUrl: 'http://localhost:3000',
+    setupNodeEvents(on, config) {
+      on('task', {
+        async resetAll() {
```
olivierobert, line 10: Adding the `All` brings more confusion as it could mean also resetting mocks or env data.

So the following would be more descriptive:

```suggestion
        async resetDatabase() {
          return resetAllDatabaseTables();
        }
```

cypress.config.ts
```diff
@@ -0,0 +1,26 @@
...
+          return resetDatabase();
+        },
+
+        async setupUser({ email, password }) {
```
olivierobert, line 14: Now I understand how this will be used after checking https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/136/files#diff-24eee1bcbd5c0632523420a2e67ab266be7c90498752546a8eb3b9021200d9e6R21

cypress.config.ts
```diff
@@ -0,0 +1,26 @@
+import { defineConfig } from 'cypress';
+import resetAllDatabaseTables from '@test/helpers/database.reset';
+import { createUser } from '@test/helpers/database.seed';
+
+export default defineConfig({
+  e2e: {
+    baseUrl: 'http://localhost:3000',
+    setupNodeEvents(on, config) {
```
tednguyendev, line 8: Unused param 👍 

```suggestion
    setupNodeEvents(on, _) {
```

tednguyendev (approved):

olivierobert (approved):

PR #136, opened 2023-09-18
[#121] [Chore] Implement UI test for login flow

### LINKED ISSUE #121: Implement UI test for the login flow
## Why

As the first step in the user flow, the login flow must be validated.

## Acceptance Criteria

- Implement UI test for the login with username + password
- Implement UI test for the login with Google

Prabeshpd:
- Close #121 

## What happened 👀

- Added cypress test for login flow
- Add the data-test attributes to login elements

## Insight 📝

- For credential login normal database interaction is used.
- For the google login request is intercepted and a token is set so that the middleware can resolve user to authenticate the user.

>[!Note] 
> - The flow of redirecting to main page is checked for google login. Since, google login is a third party service the request is intercepted
> - Token set is a encoded user as middleware layer tries to decode the user.

## Proof Of Work 📹

Passed CI pipelines


.github/workflows/test.yml
```diff
@@ -80,9 +80,11 @@ jobs:
         run: |
           touch .env.test
           echo POSTGRES_DB_CONNECTION_URL=postgresql://postgres:postgres@db:5432/ic_nextjs_test >> .env.test
+          echo NEXT_AUTH_SECRET='ENTER_NEXT_AUTH_SECRET' >> .env.test
```
tednguyendev, line 83: Is this for the [middleware](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/blob/3e39cebd1031cc994de33be9cfe9f869a68d8b59/src/middleware.ts#L8)?

Should we do just like [this](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/blob/3e39cebd1031cc994de33be9cfe9f869a68d8b59/src/config/config.ts#L33)? ✋

cypress/e2e/login/login.cy.ts
```diff
@@ -0,0 +1,84 @@
...
+  });
+
+  describe('given valid credentials', () => {
+    it('redirects user to home page', () => {
+      const email = 'dev@nimblehq.co';
+      const password = 'password';
+      cy.task('setupUser', { email, password });
+      cy.visit('/auth/login');
+      cy.findByTestId(loginPageTestIds.inputEmail).type(email);
+      cy.findByTestId(loginPageTestIds.inputPassword).type(password);
```
olivierobert, line 24: For the login flow, it makes sense to have the steps to login but for other flows, it would be best to have a command that does all of this at once (as described in https://docs.cypress.io/api/cypress-api/custom-commands#Usage)

```
cy.login();
``

File: cypress/e2e/login/login.cy.ts [hunk 3]
```diff
@@ -0,0 +1,84 @@
...
+      cy.findByTestId(loginPageTestIds.submitButton)
+        .invoke('click')
+        .hasActiveSpinner(loginAriaLabels.loading);
+
+      cy.hasToastMessage('error', 'The provided credentials are invalid.');
+    });
+  });
+});
+
+describe('Google login', () => {
```
olivierobert, line 53: Prefer to split into two files to ease management:

- `login-credentials.cs.ts`
- `login-google.cy.ts`

cypress/e2e/login/login.cy.ts
```diff
@@ -0,0 +1,84 @@
...
+  inputPassword: 'login-form-password',
+  submitButton: 'login-form-submit',
+  googleLoginButton: 'login-google',
+};
+
+const loginAriaLabels = {
+  loading: 'login-form-loader',
+};
+
+describe('Credential Login', () => {
```
olivierobert, line 12: Prefer naming UI test with `verb + component` to describe a flow/feature from the user's perspective:

```suggestion
describe('Login with Credentials', () => {
```

cypress/e2e/login/login.cy.ts
```diff
@@ -0,0 +1,84 @@
...
+        },
+        { fixture: 'OAuthToken' }
+      ).as('googleLoginResponse');
+
+      cy.intercept(
+        {
+          method: 'POST',
+          url: 'http://localhost:3000/',
+        },
+        { body: { name: 'dev nimble' } }
```
olivierobert, line 70: Why is this needed?

.github/workflows/test.yml
Prabeshpd: I am creating a separate .env.test file for it. So, added the env variable for it.

cypress/fixtures/OAuthToken.json
olivierobert, line 1: That file name casing is not consistent with the other files (which all are using kebab case):

image

It ends up creating a messy file structure.

Beyond file naming, it would be best to keep cypress-related code in `./cypress` (that's how it is expected as none of these fixtures will be used elsewhere):

image
 
See how files are named and organized in https://github.com/nimblehq/varun-web/tree/develop/cypress/fixtures

cypress/e2e/login/login.cy.ts
Prabeshpd: I have created a method for it for the other test. Since, we want to test login flow here. I have done a full test.

cypress/e2e/login/login.cy.ts
Prabeshpd: For google login we are intercepting the request and sending a custom token.
 The user would not have been inserted to database at that point.
After login as soon as we go to home page it fetches current user to display name. So, the request is intercepted to return with a response to avoid error display.

.github/workflows/test.yml
tednguyendev: Can you elaborate it 🙈

cypress/e2e/login/login-credentials.cy.ts
```diff
@@ -0,0 +1,50 @@
...
+
+describe('Login with Credentials', () => {
+  beforeEach(() => {
+    cy.task('resetDatabase');
+  });
+
+  describe('given valid credentials', () => {
+    it('redirects user to home page', () => {
+      const email = 'dev@nimblehq.co';
+      const password = 'password';
```
tednguyendev, line 19: [Ditto](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/135#discussion_r1329625535)

cypress/e2e/login/login-google.cy.ts
```diff
@@ -0,0 +1,40 @@
...
+    cy.task('resetDatabase');
+  });
+
+  describe('given valid credentials', () => {
+    it('redirects user to home page', () => {
+      cy.visit('/auth/login');
+      cy.intercept(
+        {
+          method: 'POST',
+          url: 'http://localhost:3000/api/auth/signin/google?',
```
tednguyendev, line 16: We are having some global envs for the `http://localhost:3000` right? 👀 We can use them instead of plain string like this

.github/workflows/test.yml
Prabeshpd: For testing purpose I have a separate env file with name .env.test.

-  For the config to get the env variables I am setting the env variable for next-auth. We can leave it and let the config take care of it.
-   Placing it for the clarity so that the env variables are setup for project to run.

cypress/e2e/login/login-credentials.cy.ts
tednguyendev: What about this? 👀

Prabeshpd: I have added an extra line after the page has been visited to denote arrange part is over. Is there anything more that I missed?

tednguyendev: Ah I see, thank you 👍

tednguyendev (approved):

cypress/fixtures/OAuthToken.json
Prabeshpd: The fixtures are located inside the cypress folder. 😅 .

 Regarding the naming I was following the react IC from nimble. [IC](https://github.com/nimblehq/internal-certification-react/tree/main/micky/cypress/fixtures/requests). There was camelCase and I thought this was followed in cypress fixtures. I should have followed more projects and tried to be consistent within my project. I will certainly improve on that. Thank you.

olivierobert (approved):

PR #137, opened 2023-09-19
[#123] [Chore] Implement UI test for the search result filtering flow

### LINKED ISSUE #123: Implement UI test for the search result filtering flow
## Why

The search has several options hence its flow must be validated.

## Acceptance Criteria

- Implement UI test for the search by keyword and URL + each search type
- Implement UI test for the search by keyword + each search type
- Implement UI test for the search by URL + each search type

Prabeshpd:
- Close #123 

## What happened 👀

- Added test for filtering search results
- Create a function to seed search result for user
- Moved all the selector ID to constants
- Changed the `data-testid` to `data-test-id`.

## Insight 📝

- I have moved the selector IDs to constants file inside cypress.
     -  Some of the selector ID were needed in multiple files.
     - So, have moved them to constants and used SCREAMING_SNAKE_CASE to maintain same consistency throughout project.

- Extra work of changing `data-testid` to `data-test-id` has been done. 
     - As, we use `data-test-id` as standard. Had to change the config for unit tests.
     - Since, config was changed had to change for other files.

>[!Note]
> -  Could have created separate ticket to address data-test-id but change was not much. So, did in this task.

## Proof Of Work 📹

Passed CI pipelines


cypress/e2e/searchResults/searchResultsKeyword.cy.ts
```diff
@@ -0,0 +1,74 @@
...
+  beforeEach(() => {
+    cy.task('resetDatabase');
+  });
+
+  describe('given search type is "exact"', () => {
+    it('displays valid data', () => {
+      const email = 'dev@nimblehq.co';
+      const password = 'password';
+      const queryText = 'noted';
+      cy.task('setupUserWithSearchResults', { email, password });
```
olivierobert, line 17: Not having the test data created inside each test makes it hard to reason about assertions such as:

```
cy.get('@tableRows').should('have.length', 1);
```	

The rule of thumb is not to overengineer tests. It is best to have some code repetition in tests when that helps in understanding and managing tests. 

What I recommend is the following step in each test:

```
//  Create a user. 
// Create specific search results. The test data can change for each test.

cy.login(user); // we can pass the whole object
cy.navigateTo('search-result') // custom command to navigate to each screen
```

cypress/e2e/searchResults/searchResultsKeyword.cy.ts
```diff
@@ -0,0 +1,74 @@
...
+
+      cy.get(SEARCH_RESULT_LIST_FILTER_SELECTOR_IDS.selectType).type('Exact {enter}{enter}');
+      cy.get(SEARCH_RESULT_LIST_FILTER_SELECTOR_IDS.selectScope).type('keyword {enter}{enter}');
+      cy.findByTestId(SEARCH_RESULT_LIST_FILTER_SELECTOR_IDS.inputText).type(queryText);
+      cy.findByTestId(SEARCH_RESULT_LIST_FILTER_SELECTOR_IDS.submitButton).click();
+
+      cy.findByTestId(SEARCH_RESULT_LIST_SELECTOR_IDS.tableBody).children().as('tableRows');
+
+      cy.get('@tableRows').should('have.length', 1);
+      cy.get('td[headers="keyword-column"]').should('contain', queryText);
```
Prabeshpd, line 29: Create a command for it. It is being used multiple times

cypress/constants/selectors.ts
olivierobert, line 1: While these can be shared, having a single file with tons of constants is not easy to read and manage. Since all of these seledctors are flow specific, how about moving them to each subdirectory: `login/selectors.ts`, `searchResults/selectors.ts`, etc. Then when import them we could do simplify the import to: `import * from '../../login/selectors';

cypress/constants/selectors.ts
```diff
@@ -0,0 +1,35 @@
+export const SIGN_UP_SELECTOR_IDS = {
+  inputName: 'sign-up-form-name',
+  inputEmail: 'sign-up-form-email',
+  inputPassword: 'sign-up-form-password',
+  submitButton: 'sign-up-form-submit',
+};
+
+export const SIGN_UP_ARIA_LABELS = {
+  loading: 'sign-up-form-loader',
```
olivierobert, line 9: Can't there be a single label for all screens? That spinner is the same, anyway.

```suggestion
  loading: 'application-loader',
```

That could be a shared constant 💡

cypress/constants/selectors.ts
```diff
@@ -0,0 +1,35 @@
+export const SIGN_UP_SELECTOR_IDS = {
```
olivierobert, line 1: Let's avoid plurals when possible. The following would work too:

```suggestion
export const SIGN_UP_SELECTOR = {
```

olivierobert (changes requested):

cypress/constants/selectors.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/137/commits/862724755b491f06f91679079206c56067191a8b

cypress/constants/selectors.ts
Prabeshpd: Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/137/commits/862724755b491f06f91679079206c56067191a8b

cypress/constants/selectors.ts
Prabeshpd: Have moved to individual folders. https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/137/commits/862724755b491f06f91679079206c56067191a8b

cypress/constants/selectors.ts
```diff
@@ -0,0 +1,5 @@
+export const APPLICATION_LOADER = 'application-loader';
+
+export const APPLICATION_NAV_SELECTORS = {
```
tednguyendev, line 3: 🙈
```suggestion
export const APPLICATION_NAV_SELECTOR = {
```

cypress/e2e/searchResults/searchResultsKeyword.cy.ts
Prabeshpd: Makes sense. I was thinking the same. But got deviated 😅 thinking to make a global setup. Makes more sense to put control to test suites. Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/137/commits/8b58f43b403135cbf04f2cbc8ae31bbcc9662b97

cypress/e2e/signUp/selectors.ts
```diff
@@ -0,0 +1,10 @@
+export const SIGN_UP_SELECTOR = {
+  inputName: 'sign-up-form-name',
+  inputEmail: 'sign-up-form-email',
+  inputPassword: 'sign-up-form-password',
+  submitButton: 'sign-up-form-submit',
+};
+
+export const SIGN_UP_ARIA_LABEL = {
```
tednguyendev, line 8: Don't we use this? 👀

cypress/e2e/login/selectors.ts
```diff
@@ -0,0 +1,10 @@
+export const LOGIN_SELECTOR = {
+  inputEmail: 'login-form-email',
+  inputPassword: 'login-form-password',
+  submitButton: 'login-form-submit',
+  googleLoginButton: 'login-google',
+};
+
+export const LOGIN_ARIA_LABEL = {
```
tednguyendev, line 8: Ditto ☝️

cypress/e2e/searchResults/searchResultsUrlWithKeyword.cy.ts
tednguyendev, line 1: Maybe `searchResultsWithUrlAndKeyword.cy.ts`

src/components/SearchResultList/FilterForm.tsx
```diff
@@ -53,6 +53,7 @@ const FilterForm = (props: FilterFormProps) => {
         >
           <Select
             aria-label="Search Type"
+            id="list-search-result-form-type"
```
tednguyendev, line 56: Why do we use `id` instead of `data-test-id` 👀

Prabeshpd: React Selector don't embed element with data-select-id apparently. Tried it. 😅

cypress/e2e/login/selectors.ts
Prabeshpd: OOps. Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/137/commits/7873beb9a686d7b5bb4cf78aa834b3a9574deb49

tednguyendev (approved):

cypress/e2e/searchResults/searchResultsKeyword.cy.ts
```diff
@@ -0,0 +1,120 @@
...
+        {
+          userId,
+          keyword: 'noted',
+        },
+        {
+          userId,
+          keyword: 'litter',
+        },
+      ];
+      cy.task('setupSearchResults', searchResultPayload);
```
olivierobert, line 31: Readability improvement suggestion:

```
const queryText = 'text_to_find';
const [ userId, email, password ]  = [1,  'dev@nimblehq.co'; 'password' ]
const searchResultPayload = [...]

cy.task('setupUser', { id: userId, email, password });
cy.task('setupSearchResults', searchResultPayload);

cy.login({ email, password });
...
```

The first three lines are part of the Arrange section, and the rest are steps for the Act section. Plabing the query text at the top allows to making of the search result payload.

cypress/e2e/searchResults/searchResultsKeyword.cy.ts
```diff
@@ -0,0 +1,120 @@
...
+          userId,
+          keyword: 'no/de',
+        },
+        {
+          userId,
+          keyword: 'noted',
+        },
+        {
+          userId,
+          keyword: 'litter',
```
olivierobert, line 28: This extra search result is not really needed. We need at most 2 search results to validate the test. So lets create the minimum test data.

olivierobert (changes requested):
Since the tests are pretty long, it is best to adjust them for readability. The suggestion I have raised should be applied to all tests.

cypress/support/commands.ts
```diff
@@ -2,12 +2,31 @@
...
+Cypress.Commands.add('login', (user: any) => {
+  cy.visit('/auth/login');
+  cy.findByTestId(LOGIN_SELECTOR.inputEmail).type(user.email);
+  cy.findByTestId(LOGIN_SELECTOR.inputPassword).type(user.password);
+
+  cy.findByTestId(LOGIN_SELECTOR.submitButton).click();
+});
+
+Cypress.Commands.add('navigateTo', (navId: string) => {
+  cy.findByTestId(navId).click();
```
olivierobert, line 27: This comamnd could be improved further by handling all possible pages (there are not many):

```
import { APPLICATION_NAV_SELECTOR } from '../../constants/selectors';

Cypress.Commands.add('navigateTo', (pageName: string) => {
   switch(pageName) {
      case 'searchResult': 
         cy.navigateTo(APPLICATION_NAV_SELECTOR.searchResult);
     //...

     default:
       // throw error if page not supported
   }
}
```

Then it would be easier to use: `cy.navigateTo('searchResult')` 🎉

cypress/e2e/searchResults/searchResultsKeyword.cy.ts
Prabeshpd: Rearranged it as per suggestion. Thank you. Fixed in https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/137/commits/7679aee28952f75573eac6126ae38928270ac41e

olivierobert (approved):

Issue #126, opened 2023-09-08
[Chore] Set up Docker compose to deploy the application


olivierobert:
## Why

To easily deploy the application, it must be fully containerized.

## Acceptance Criteria

- Set up a Docker Compose file for deployment: `docker-compose.yml`
- Set up service for the worker.
- Set up a service for the application. 
- Do NOT set up a service for the database as a third-party service will be used.

Prabeshpd:
Since for worker we only require the redis instance. I believe we will use  a hosted service. We would not have service for worker. Is that OK?

PR #141, opened 2023-09-22
[#23] [Chore] Set up Docker compose for local development

### LINKED ISSUE #23: [Chore] Set up Docker compose for local development
## Why
Setting up a Docker Compose file will help other developers start the project easily and start the development. 

## Acceptance Criteria

- Set up a Docker Compose file for local development: `docker-compose.dev.yml`
- Set up service for the database.
- Set up service for the worker.
- (Optional) Set up a service for the application. 
- Add a Makefile to start the containers (see https://makefiletutorial.com/). The command could be: `make start`

Prabeshpd:
- Close #23 

## What happened 👀

- Added a config for redis connection
- Used redis connection on worker
- Updated workflow for redis
- Added docker file for application
- Added docker compose for development
- Added makefile

## Insight 📝

- Redis connection is passed for bullmq workers
     - SInce, we are using docker. Application will not have access to docker containers through localhost
     - Bullmq directly looked for localhost connection
     - For this we are adding a redis connection
     - For redis connection singleton pattern is used.
 - Added a docker-compose file
- Added a make file
> [!Note]
> - I have not used the docker image of application on development. Since, we want the latest changes of development to visible. 
> - We can mount the volume of directory source to docker source directory. I have avoided that.
> - I have used the local server and docker instance of redis and postgres.

## Proof Of Work 📹

https://github.com/nimblehq/ic-nextjs-prabesh-dahal/assets/39412015/b6762ed3-6fe4-402f-a95a-fbd862018d15


Dockerfile
```diff
@@ -0,0 +1,56 @@
+FROM node:18-bullseye-slim AS base
```
Prabeshpd, line 1: For the spacing :
I did not find any specifications on compass for dockerfile

- I have added a line break to separate the docker image layers.
- I have tried to group the same command layer and added extra line to separate them.
- As, I have built a multi step docker I have added two lines between each step.

tailwind.config.js
```diff
@@ -5,14 +5,14 @@ module.exports = {
     './src/components/**/*.{js,ts,jsx,tsx,mdx}',
     './src/app/**/*.{js,ts,jsx,tsx,mdx}',
   ],
+  darkMode: 'class',
```
olivierobert, line 8: This change is outside the scope of change for this PR 🤔 It is best to not buncle this in the same set of changes (I also do know why this is needed or what this will do).

.env.example
```diff
@@ -8,3 +8,10 @@ NEXT_AUTH_SECRET=
 
 # Request  in Next.JS > 13 does not support the query attribute, hence requires a fully qualified URL
 TEST_APP_HOST=
+
+# Docker env variables
```
olivierobert, line 12: It is confusing to add this description as passing env variables to a Docker container has its own syntax (see https://vsupalov.com/docker-arg-env-variable-guide/#setting-env-values)

In addition, I do not see these env variables being used in the codebase 🤔 So why are these PG related env needed?

Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+
+# Install required dependecies
+FROM base as dependencies
+
+#Puppeteer installation
```
olivierobert, line 6: This comment is rather misleading as the code skips the installation of Puppeteer (but the tile says otherwise)

```suggestion
# Disable Puppeteer installation
```

Also, why disabling it? Wouldn't it be required for the PDF export?

Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+    apt-get install chromium -y --no-install-recommends && \
+    rm -rf /var/lib/apt/lists/*
+
+WORKDIR /app
+
+COPY package.json yarn.lock ./
+
+RUN yarn --frozen-lockfile
+
+
```
olivierobert, line 23: No convention on that front, but the extra line break can be removed. The comment block is sufficient to delimit multi-stage build

Dockerfile
olivierobert, line 1: Why is this file added in this PR? Shouldn;t this be for #126? Since this PR is for local development and the app is not run in a container, there is no need for such a file yet.

Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+ENV NODE_ENV production
+
+WORKDIR /app
+
+RUN addgroup --system --gid 1001 nodejs && \
+    adduser --system --uid 1001 nextjs
+RUN mkdir .next && \
+    chown nextjs:nodejs .next
+
+COPY --from=builder --chown=nextjs:nodejs /app .
```
olivierobert, line 48: Wouldn't that copy also all subdirectories in `build`?

Makefile
```diff
@@ -0,0 +1,13 @@
...
+
+.PHONY: dev-start dev-stop test
+
+dev-start:
+	docker compose -f $(dev_compose_file) up -d
+	yarn migrate:dev
+	yarn dev
+
+dev-stop:
+	docker compose -f $(dev_compose_file) down -v
```
olivierobert, line 13: For local dev, there would be no need to remove each time, so the following would suffice:

```suggestion
	docker compose -f $(dev_compose_file) stop -v
```

src/config/redis.ts
```diff
@@ -0,0 +1,34 @@
...
+  host: process.env.REDIS_HOST || 'localhost',
+  port: Number(process.env.REDIS_PORT || 6379),
+  password: process.env.REDIS_PASSWORD || '',
+};
+
+export default class Redis {
+  private static instance: IORedis;
+
+  private constructor() {
+    throw new NotImplementedError({ message: 'Class can not be instantiated' });
```
olivierobert, line 21: Having a class that cannot be instantiated is unexpected. Two options: 1) move the  code for `getInstance` to the constructor method or 2) remove the constructor method altogether. 

Here is below, an example from the easyHotel APi codebase:

src/workers/queueScraper.worker.ts
```diff
@@ -1,15 +1,23 @@
 import { Worker, Queue } from 'bullmq';
 
+import Redis from '@/config/redis';
 import { AsyncProcess } from '@/constants/enum';
 import { BullQueue } from '@/lib/worker';
 
 import buildScraper from './queueScraper';
 
-const scrapeBullQueue = new Queue(AsyncProcess.QUEUE_SCRAPE);
+const redis = Redis.getInstance();
```
olivierobert, line 9: Let's be more descriptive, especially as there is another variable close to this one (`Redis`):

```suggestion
const redisConnection = Redis.getInstance();
```

.env.example
Prabeshpd: It's needed for the docker-compose file to add environment. Since, there are few environment we can directly inject to compose file. Makes sense will remove it.

Dockerfile
Prabeshpd: Yes, but next application build has lots of internal folders which if we do not copy will not render the CSS images.
Had a lot of copy statements lines so, ended up copying all the folders.

Dockerfile
Prabeshpd: It installs chromium under the hood. The environment variable for PUPPETEER_SKIP_CHROMIUM_DOWNLOAD is to skip the latest version of chromium. But I guess it can be removed.

src/config/redis.ts
Prabeshpd: I can not see the image. 😅 . Have removed the constructor for now.

Prabeshpd:
Have moved the docker file to https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/143

src/workers/queueScraper.worker.ts
Prabeshpd: Fixed.

Dockerfile
Prabeshpd: Got it.

tailwind.config.js
Prabeshpd: Removed it.

Makefile
Prabeshpd: Fixed

tailwind.config.js
```diff
@@ -9,10 +9,9 @@ module.exports = {
     extend: {
       backgroundImage: {
         'gradient-radial': 'radial-gradient(var(--tw-gradient-stops))',
-        'gradient-conic':
-          'conic-gradient(from 180deg at 50% 50%, var(--tw-gradient-stops))',
+        'gradient-conic': 'conic-gradient(from 180deg at 50% 50%, var(--tw-gradient-stops))',
```
Prabeshpd, line 12: Code format revamped

src/config/redis.ts
olivierobert: I have reuploaded the image

src/config/redis.test.ts
```diff
@@ -0,0 +1,22 @@
...
+import Redis from './redis';
+
+describe('Redis:', () => {
+  let redis: IORedis;
+
+  afterAll(() => {
+    redis.disconnect();
+  });
+
+  describe('given "getInstance" is called multiple times', () => {
```
tednguyendev, line 15: Should we also test the case that: When there is no instance, it init a `IORedis` 👀

README.md
```diff
@@ -38,6 +38,18 @@ The application is bootstrapped with [`create-next-app`](https://github.com/verc
...
    ```
 
+## Docker Compose
+
+1. Start Development Server
+
+   ```bash
+   cp .env.example .env
+
+   make dev-start
```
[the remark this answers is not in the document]
Prabeshpd: You may need to add POSTGRES_DB_CONNECTION_URL in .env.test. That usually depends on your configuration. So, have added empty string in .env.example.

I will add a extra information to make sure to add the URL in Readme.

Makefile
```diff
@@ -0,0 +1,12 @@
+dev_compose_file := docker-compose.dev.yml
+application_container_test := application_test
```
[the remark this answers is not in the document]
Prabeshpd: Removed it.

README.md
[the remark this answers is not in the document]
Prabeshpd: Updated the Readme

src/config/redis.test.ts
Prabeshpd: Oh. you mean to ensure that the instance is of IORedis. Makes sense. Have updated it.

README.md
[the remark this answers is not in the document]
tednguyendev: Yes it works now.

1 point is that I think we should move [this part](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/141/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R9-R13) to under [this part](https://github.com/nimblehq/ic-nextjs-prabesh-dahal/pull/141/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R15-R19).

As in the current order: If we add the `POSTGRES_DB_CONNECTION_URL` to `.env`, then `cp .env.example .env`, the `.env` will be override by the `.env.example`, make it not has the `POSTGRES_DB_CONNECTION_URL` anymore 👍

README.md
```diff
@@ -2,6 +2,24 @@
 
 The application is bootstrapped with [`create-next-app`](https://github.com/vercel/next.js/tree/canary/packages/create-next-app).
 
+## Docker Compose
+
+- ### Start Development Server
```
tednguyendev, line 7: How about just

```suggestion
## Docker

```

tednguyendev (approved):

README.md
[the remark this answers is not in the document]
Prabeshpd: Done.

README.md
Prabeshpd: I have added readme for test part as well. So, it will help us to differentiate between local development and testing.

.env.example
```diff
@@ -8,3 +8,7 @@ NEXT_AUTH_SECRET=
 
 # Request  in Next.JS > 13 does not support the query attribute, hence requires a fully qualified URL
 TEST_APP_HOST=
+
+REDIS_HOST=
```
olivierobert, line 12: It would be better to restructure the env as now it seems that Redis is part of the test env variable:

```
POSTGRES_DB_CONNECTION_URL=
REDIS_HOST=
REDIS_PORT=
REDIS_PASSWORD=

REACT_BASE_ENV_URL=
NEXT_AUTH_SECRET=

OAUTH_GOOGLE_CLIENT_ID=
OAUTH_GOOGLE_CLIENT_SECRET=

# Request  in Next.JS > 13 does not support the query attribute, hence requires a fully qualified URL
TEST_APP_HOST=
```

Also, could `TEST_APP_HOST=` be moved to an `.env.test`? Not a blocker but it could be a good move.

olivierobert (approved):
Just a minor suggestion for you to consider. Up to you to decide to add it in or not.

.env.example
Prabeshpd: Updated it.

